PR #77, opened 2020-05-31
Deprovining instance removes ALIAS records

tammersaleh:

Also, improved local dev docker process:
- Added more files to .dockerignore
- `./dev` now runs commands through bash -c
- That was in order to run the `new-migration` command as a single `docker run`.

Also:
- Fixed issue where Enums were being stored as integers instead of strings.

[Closes #722]

## Security considerations

None


tammersaleh:
@bengerman13 created this PR to let you know that I implemented a small part of your refactoring here:  Removed `Operation.queue_tasks()`, and instead called `tasks.queue_all_provision_tasks_for_operation()` inside `API`.  This will likely be confusing as you rebase your branch, so figured I'd mention it.

PR #145, opened 2020-08-03
Clean up certs

bengerman13:

- Remove old certificate from ALB, IAM, and database after rotating certificates.

## Security considerations

None


bengerman13:
Recommended reading order:
- tests/integration/test_cdn_*
- broker/tasks/pipelines.py
- broker/tasks/iam.py
- broker/models.py
- tests/integration/test_alb_*
- broker/tasks/alb.py

migrations/versions/ad4dff9923c0_previous_listener_data_for_alb.py
```diff
@@ -0,0 +1,36 @@
+"""previous listener data for ALB
```
bengerman13, line 1: This file it auto-generated

tests/lib/fake_iam.py
```diff
@@ -26,7 +26,7 @@ def expect_upload_server_certificate(
                 "ServerCertificateId": "FAKE_CERT_ID_XXXXXXXX",
                 "Path": path,
                 "ServerCertificateName": name,
-                "Arn": f"arn:aws:iam::000000000000:server-certificate{path}/{name}",
+                "Arn": f"arn:aws:iam::000000000000:server-certificate{path}{name}",
```
ccostino, line 29: The forward slash isn't needed here?  I'm guessing it's included in one of the variables being interpolated in?

broker/tasks/alb.py
```diff
@@ -91,3 +96,34 @@ def remove_certificate_from_alb(operation_id, **kwargs):
...
+
+    time.sleep(int(config.DNS_PROPAGATION_SLEEP_TIME))
+
+    if service_instance.previous_alb_listener_arn is not None:
+        alb.remove_listener_certificates(
+            ListenerArn=service_instance.previous_alb_listener_arn,
+            Certificates=[
+                {"CertificateArn": remove_certificate.iam_server_certificate_arn}
+            ],
+        )
```
ccostino, line 124: I'm just making sure I understand the flow correctly here - this is removing the previous certificate from the ALB, right?

broker/tasks/iam.py
```diff
@@ -79,3 +80,34 @@ def delete_server_certificate(operation_id: str, **kwargs):
...
+            Certificate.id != service_instance.current_certificate_id,
+        )
+    ).all():
+        try:
+            iam.delete_server_certificate(
+                ServerCertificateName=certificate.iam_server_certificate_name
+            )
+        except iam_commercial.exceptions.NoSuchEntityException:
+            pass
+        db.session.delete(certificate)
```
ccostino, line 111: I was going to ask if this was doable within the loop as you'd be modifying a list/set of elements in place, but that's not actually the case - you're queuing things up for a transaction and committing below, so it should be fine! :-)

broker/tasks/pipelines.py
```diff
@@ -282,6 +276,17 @@ def queue_all_alb_update_tasks_for_operation(operation_id, correlation_id):
...
+        .then(
+            alb.remove_certificate_from_previous_alb,
+            operation_id,
+            correlation_id=correlation_id,
+        )
+        .then(
+            iam.delete_previous_server_certificate,
+            operation_id,
+            correlation_id=correlation_id,
+        )
```
ccostino, line 289: Based on the flow here, it's wait until the ALB is ready to be modified, then remove the certificate from the ALB, then delete the certificate itself, correct?

tests/lib/fake_iam.py
bengerman13: yeah, this was also a nit-picky thing I just stumbled on while testing. No current functionality makes any assumptions about what the ARN looks like, so this really just makes the error look more realistic if/when your tests fail

broker/tasks/alb.py
bengerman13: correct

broker/tasks/pipelines.py
bengerman13: right - we need to wait for the route53 changes to propagate so we're sure we can remove the cert from the old ALB. Then we want to remove the cert from the ALB before removing it from IAM. (we actually could remove it from IAM first, then ALB second, but it's not recommended and logs some warnings, in addition to not making any sense to do)

migrations/versions/ad4dff9923c0_previous_listener_data_for_alb.py
ccostino: Gotta love migration frameworks. :-)

tests/lib/fake_iam.py
ccostino: Awesome, thanks!

broker/tasks/pipelines.py
ccostino: Fair enough!  And you're right, it doesn't make sense to do it in reverse order.

Thank you for explaining, I appreciate it!

broker/tasks/iam.py
bengerman13: right - the `db.session.delete()` says "queue up a delete operation for this object" then `db.session.commit()` does the actual delete.

ccostino: *(tents fingers)* Excellent.

PR #147, opened 2020-08-03
add more docs

bengerman13:

- add more docs
- remove some no-op tests

## Security considerations

None

### DIFF
--- a/docs/operational.md
+++ b/docs/operational.md
@@ -1,0 +1,66 @@
+# Operational docs
+## tips and tricks for running this thing IRL
+
+
+### Things to keep in mind
+
+Huey pipelines are linked-list tasks. This means that when Huey pulls a task out of the queue,
+and that task is part of a pipeline, the whole pipeline is popped from the queue. Because of this,
+if the worker consuming a task terminates without pushing the pipeline back to the queue, Huey loses
+track of the pipeline. This lead to the creation of the `scan_for_stalled_pipelines` job, which runs
+on a cron schedule. It looks for pipelines that have not been updated in long enough that it appears
+they're not running. When such pipelines are detected, they're re-enqueued from the start.
+
+### Stopping pipelines by hand
+
+Sometimes, a pipeline might be running, and you may want to make it stop. You can do this 
+by connecting to the database and updating the Operation's `canceled_at` field:
+
+```
+sql> UPDATE operation SET canceled_at = now() WHERE id = <operation_id>;
+```
+
+### Restarting a pipeline
+
+Sometimes it's useful to restart a pipeline manually, for instance if a pipeline is failing
+on a step, and you know rerunning a previous step is likely to fix it.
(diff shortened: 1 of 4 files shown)


docs/operational.md
```diff
@@ -0,0 +1,66 @@
...
+```
+sql> SELECT id FROM operation WHERE
+        service_instance_id = <service_instance_id> AND
+        state = 'in progress' AND
+        canceled_at IS NULL;
+```
+
+Example: Someone, for some reason, ran `cf purge-service-instance <service_instance>`,
+but we need to delete the service instance from the broker so the AWS resources get cleaned up.
+All we need to do is create a record in the `operation` table and wait:
```
ccostino, line 54: In other words, someone went to remove a service and did so directly instead of queuing up a job to take care of all the steps?  And by following these instructions, this'll cause the `scan_for_stalled_pipelines` job to pick up on that and take care of everything else that would have been done?

docs/operational.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+### Things to keep in mind
+
+Huey pipelines are linked-list tasks. This means that when Huey pulls a task out of the queue,
+and that task is part of a pipeline, the whole pipeline is popped from the queue. Because of this,
+if the worker consuming a task terminates without pushing the pipeline back to the queue, Huey loses
+track of the pipeline. This lead to the creation of the `scan_for_stalled_pipelines` job, which runs
+on a cron schedule. It looks for pipelines that have not been updated in long enough that it appears
+they're not running. When such pipelines are detected, they're re-enqueued from the start.
```
ccostino, line 12: Based on reading this, the pipeline itself is not lost in this situation, correct?  There's a means of retrieving it again to place it back in the queue?

bengerman13: sort of - there are two persistence models here:
- Redis, used only by Huey
- Postgres, used directly by the broker code and the pipelines we feed to Huey

Redis has the pipeline itself - the data about what huey tasks to run in what order, and what tasks have been run already, etc.
Postgres has the data about the service instance - its current and desired states, and other such data.

When we run into this issue, the data from redis is gone entirely, but the data in postgres is still there.  The `scan_for_stalled_pipelines` job uses the `operation` model in postgres to tell Huey to create a new pipeline in Redis.

ccostino: Okay, that sounds like it answers my question; a pipeline can be reconstructed to be re-run at a later point in time, even though the previous pipeline is gone (which is to be expected)?

docs/operational.md
bengerman13: yeah, this is kind of a contrived example. The important scenario in this example is that the service instance needs to be deleted, and for some reason we can't tell CAPI to do it.

ccostino: Ah I see, that makes sense.  Thank you!

docs/operational.md
ccostino: The answer is yes. :-)

Issue #149, opened 2020-08-04
on update, don't get new cert unless it's needed


bengerman13:
If the domains don't change on an update, don't get a new certificate.

PR #150, opened 2020-08-04
don't update ALBs when nothing has changed

bengerman13:

-  when the domains are the same and the instance type is ALB, don't do anything on update

## Security considerations

None

### DIFF
--- a/broker/api.py
+++ b/broker/api.py
@@ -260,12 +260,14 @@ class API(ServiceBroker):
         domain_names = [
             d.strip().lower() for d in params.get("domains", "").split(",") if len(d)
         ]
+        noop = True
         if len(domain_names):
             self.logger.info("validating CNAMEs")
             validators.CNAME(domain_names).validate()
 
             self.logger.info("validating unique domains")
             validators.UniqueDomains(domain_names).validate(instance)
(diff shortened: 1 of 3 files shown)


bengerman13:
this is part of #149 - the CDN part is a bit more involved, so I'll address it in another PR

broker/api.py
```diff
@@ -343,6 +346,9 @@ def update(
             queue = queue_all_cdn_update_tasks_for_operation
         else:
             queue = queue_all_alb_update_tasks_for_operation
+        if noop:
+            return UpdateServiceSpec(False)
```
ccostino, line 350: If `noop` is `True` then run this?  What does `UpdateServiceSpec` do?  I see you're explicitly passing in `False` compared to other similar calls here, so I'm assuming this is indicating that nothing should take place but a record of the action needs to still be made?

bengerman13: `UpdateServiceSpec` is the class the osbapi package uses to represent the return value of an update call. The parameter controls whether or not the return indicates the update is async. We have no work to do, so this is ~not async~ _synchronous_.
The upshot is that `UpdateServiceSpec(False)` will return a 200, and `UpdateServiceSpec(True)` will return a 202

ccostino: Excellent, thanks!

PR #152, opened 2020-08-05
Refactor cdn update parsing

bengerman13:

- refactor update and provisioning logic for CDN to prep for no-op checks and to share more logic
- disable random nonce-rejection from pebble test server. The thinking here is that we have sufficient retries in production, but the way we retry in production isn't reflected in our local tests, so injecting the failures into local testing just makes tests flaky

## Security considerations

None

### DIFF
--- a/broker/api.py
+++ b/broker/api.py
@@ -21,6 +21,8 @@ from openbrokerapi.service_broker import (
     UpdateServiceSpec,
 )
 from sap import cf_logging
+from sqlalchemy.dialects import postgresql
+
 
 from broker import validators
 from broker.extensions import config, db
@@ -136,30 +138,14 @@ class API(ServiceBroker):
             )
             instance.cloudfront_origin_path = params.get("path", "")
             instance.route53_alias_hosted_zone = config.CLOUDFRONT_HOSTED_ZONE_ID
-            forward_cookies = params.get("forward_cookies", None)
-            if forward_cookies is not None:
-                forward_cookies = forward_cookies.replace(" ", "")
-                if forward_cookies == "":
-                    instance.forward_cookie_policy = (
-                        CDNServiceInstance.ForwardCookiePolicy.NONE.value
-                    )
-                elif forward_cookies == "*":
-                    instance.forward_cookie_policy = (
-                        CDNServiceInstance.ForwardCookiePolicy.ALL.value
-                    )
-                else:
-                    instance.forward_cookie_policy = (
-                        CDNServiceInstance.ForwardCookiePolicy.WHITELIST.value
-                    )
-                    instance.forwarded_cookies = forward_cookies.split(",")
-            forwarded_headers = params.get("forward_headers", None)
-            if forwarded_headers is None:
-                forwarded_headers = []
-            else:
-                forwarded_headers = forwarded_headers.replace(" ", "")
-                forwarded_headers = forwarded_headers.split(",")
-            if params.get("origin") is None:
+            forward_cookie_policy, forwarded_cookies = parse_cookie_options(params)
+            instance.forward_cookie_policy = forward_cookie_policy
+            instance.forwarded_cookies = forwarded_cookies
+            forwarded_headers = parse_header_options(params)
+            if instance.cloudfront_origin_hostname == config.DEFAULT_CLOUDFRONT_ORIGIN:
                 forwarded_headers.append("HOST")
+            forwarded_headers = normalize_header_list(forwarded_headers)
+
             instance.forwarded_headers = forwarded_headers
             if params.get("insecure_origin", False):
                 if params.get("origin") is None:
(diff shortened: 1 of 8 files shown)


broker/api.py
```diff
@@ -278,59 +263,35 @@ def update(
...
-                    instance.cloudfront_origin_hostname
-                    == config.DEFAULT_CLOUDFRONT_ORIGIN
-                ):
-                    forwarded_headers.append("HOST")
-                instance.forwarded_headers = forwarded_headers
+                forwarded_headers = parse_header_options(params)
+            else:
+                # this is a weird way to do this, but if we _don't_ do this, sqlalchemy doesn't
+                # recognize that instance.forwarded_headers gets changed
+                forwarded_headers = [*instance.forwarded_headers]
```
bengerman13, line 288: I really want a more readable thing here. The best I can come up with is something like:
```python
def copy_but_dont_tell_sqlalchemy(list_):
    new_list = [*list_]
    return new_list
```

broker/api.py
```diff
@@ -278,59 +263,34 @@ def update(
...
-                    # to trick the logic below into updating it.
-                    params["forward_headers"] = params.get(
-                        "forward_headers", []
-                    ).append("HOST")
+                if params["origin"]:
+                    origin_hostname = params["origin"]
                 else:
-                    instance.cloudfront_origin_hostname = params["origin"]
+                    origin_hostname = config.DEFAULT_CLOUDFRONT_ORIGIN
+                instance.cloudfront_origin_hostname = origin_hostname
```
ccostino, line 270: For this check and the one below, is it only necessary to set the `origin_hostname` or `cloudfront_origin_path` if those keys exist in the `params` in the first place, or should they always have a value?

If it's only if the keys exist then this is fine!  If there must always be something though, could these checks be consolidated slightly, such as:

```python
# params.get() will return None by default if the dict key doesn't exist: https://docs.python.org/3/library/stdtypes.html#dict.get
origin_hostname = params.get('origin')

if origin_hostname is None:
    origin_hostname = config.DEFAULT_CLOUDFRONT_ORIGIN

instance.cloudfront_origin_hostname = origin_hostname
```

Actually, might this be possible anyway even if the initial key checks need to stay as is?

broker/api.py
```diff
@@ -278,59 +263,34 @@ def update(
...
-                if (
-                    instance.cloudfront_origin_hostname
-                    == config.DEFAULT_CLOUDFRONT_ORIGIN
-                ):
-                    forwarded_headers.append("HOST")
-                instance.forwarded_headers = forwarded_headers
+                forwarded_headers = parse_header_options(params)
+            else:
+                # .copy() so sqlalchemy recognizes the field has changed
+                forwarded_headers = instance.forwarded_headers.copy()
```
ccostino, line 287: 👏 - sorry this didn't cross my mind earlier when we were debugging!  I'm assuming just the [shallow copy](https://docs.python.org/3/library/copy.html#copy.copy) works here and there isn't the need of a [deep copy](https://docs.python.org/3/library/copy.html#copy.deepcopy)?

Thanks, @spgreenberg and @onelittlebecca, as you helped uncover what was going on here! :-)

broker/api.py
```diff
@@ -382,3 +341,40 @@ def unbind(
...
+            forward_cookie_policy = CDNServiceInstance.ForwardCookiePolicy.NONE.value
+            forwarded_cookies = []
+        elif forward_cookies == "*":
+            forward_cookie_policy = CDNServiceInstance.ForwardCookiePolicy.ALL.value
+            forwarded_cookies = []
+        else:
+            forward_cookie_policy = (
+                CDNServiceInstance.ForwardCookiePolicy.WHITELIST.value
+            )
+            forwarded_cookies = forward_cookies.split(",")
```
ccostino, line 360: Would it make sense to set `forwarded_cookies = []` at the start of the method (along with `forward_cookies`) and only modify if need be, which only appears to be in this particular case?  It looks like it's needed and used regardless all the way up to the `return` statement.

broker/api.py
bengerman13: yeah, shallow copy should be fine here - nothing else can/should be messing with the list members, and all we really need to accomplish is getting sqlalchemy to see the reference change

broker/api.py
bengerman13: yeah, it's only if the key exists. The OSBAPI says we should not change anything that is not in the update parameters, so we leave it as-is otherwise. We then check for the truthiness of the origin. The intent here is that a user could update to not use a custom origin by passing `"origin": null` or `"origin": ""`

broker/api.py
bengerman13: I think I like this way better, as this is basically a set of finite options (all, none, allowlist) and not building a list of features

broker/api.py
ccostino: Ah okay, so leave it alone unless the user has explicitly done something different here.  Thanks!

broker/api.py
ccostino: Fair enough!

PR #160, opened 2020-08-10
use a useable python version

bengerman13:

-  use a useable python version. 3.8.5 exists in dev and staging, but not production

## Security considerations

None


onelittlebecca:
:-)

PR #161, opened 2020-08-10
force operation update on each step

### LINKED ISSUE #159: Failed pipelines seem not to abort
Over the weekend, we received 15 failed pipeline emails for the same operation at different steps. It seems the pipeline failure logic needs to be updated.

bengerman13:

- Add a test to make sure that fail pipelines are totally abandoned (this turned out to be unnecessary, but why get rid of a test I already wrote?)
- Mark `step_description` as updated so updated_at gets bumped when we fail a step.
- fixes #159 

I am not confident this is the _best_ way to do this, but I think it works.

I believe what's happening here is this:
- pipeline reaches step that can never succeed, and fails, entering retry loop
- after it's been in the retry loop long enough, the stalled pipeline scanner detects it, and enqueues another pipeline for the same operation
- this happens again until the first pipeline hits the retry timeout. At this point, the operation's state changes, so the stalled pipeline scanner doesn't find it anymore. We also get an email at this point
- the rest of the pipelines burn through their retries, then send us staggered emails as they finish

This solution attempts to work around this by forcing sqlalchemy to make a SQL `UPDATE` on the operation's `step_description`, even if it hasn't changed. 

Other options I've considered, but am not crazy about:
- add a `pre_execute` or `post_execute` that does the same thing. This makes it harder to forget to do this, but it would have to make an assumption about the arguments passed to the task
- change the staleness check on stalled pipelines to be longer than the longest a task can get stuck in a retry loop. This would mean stale pipelines could be stuck for four hours, which seems too long to me.
- a bigger, crazier option I've considered is to make all our tasks into callable classes, so things like this can be more automated, but it would be significant work for probably not much benefit. I also suspect this would interfere with how Huey works (I think it actually maps our functions to the `__call__` method of other classes under the hood, so stuff might get weird.)

## Security considerations

None


broker/tasks/alb.py
```diff
@@ -2,6 +2,7 @@
 import time
 
 from sqlalchemy import and_
+from sqlalchemy.orm.attributes import flag_modified
```
ccostino, line 5: Interesting!  I [looked this up](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.attributes.flag_modified) and wonder if this would've helped with the other issue we were seeing with the list value not being set, though I suspect the fix we made was still correct given Python's way of handling variable assignments.

tests/integration/test_failing_tasks.py
```diff
@@ -18,6 +18,17 @@ def test_operations_marked_failed_after_failing(
     def no_retry_task(operation_id):
         raise Exception()
 
+    @huey.task()
+    def semaphore_task(operation_id):
+        # this task deletes our operation
+        # so tests will fail if the task failure logic
+        # doesn't abort the rest of the pipeline
+
+        with no_context_app.app_context():
```
ccostino, line 27: 🤔 "with no context app, call the app context..." Reading this out loud was entertaining. 😄

tests/integration/test_iam_idempotent.py
```diff
@@ -1,4 +1,5 @@
 from datetime import date
+import time
```
ccostino, line 2: Is the `time` import still needed here?  I don't see any specific reference to it within this file.

broker/tasks/alb.py
bengerman13: I think it would've worked, yeah.

I don't know which I'd call the "right" way. This way, IMO, seems more obvious as to _why_ we're doing it. The other way seems more obvious about _what_ we're doing.

tests/integration/test_iam_idempotent.py
bengerman13: :doh: I'd added a time.sleep when I was trying to debug this. Good catch

bengerman13: fixed

ccostino:
@bengerman13 looks okay now, just needs to be updated with the main branch!

PR #162, opened 2020-08-10
Custom 404

### LINKED ISSUE #148: Support custom 404 page
Federalist often uses custom 404 pages on their sites, and this is currently configured by hand. We should allow configuring a custom 404 page using the broker.

bengerman13:

- Add support for custom error responses in cloudfront

Fixes #148 

## Security considerations

None

### DIFF
--- a/broker/tasks/cloudfront.py
+++ b/broker/tasks/cloudfront.py
@@ -272,11 +272,11 @@ def update_distribution(operation_id: str, **kwargs):
     service_instance = operation.service_instance
     certificate = service_instance.new_certificate
 
-    config = cloudfront.get_distribution_config(
+    config_response = cloudfront.get_distribution_config(
         Id=service_instance.cloudfront_distribution_id
     )
-    etag = config["ETag"]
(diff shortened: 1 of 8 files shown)


broker/tasks/cloudfront.py
```diff
@@ -265,30 +275,33 @@ def update_distribution(operation_id: str, **kwargs):
     config = cloudfront.get_distribution_config(
         Id=service_instance.cloudfront_distribution_id
     )
-    config["DistributionConfig"]["ViewerCertificate"][
+    etag = config["ETag"]
+    config = config["DistributionConfig"]
```
ccostino, line 279: I'm just making sure that this intentional with re-using the variable name after it's initially set by calling the `cloudfront` API.

bengerman13: yeah, it was intentional, but I could see how it might not be great. I'll push a fix real quick

bengerman13: Updated :)

PR #181, opened 2020-10-01
Allow arrays for domains parameter

bengerman13:

- We've had a few users bitten by this - arrays are the obvious choice for the domain parameter, but we went with comma-separated strings initially to be consistent with the old brokers. This change allows either

## Security considerations

None

### DIFF
--- a/broker/api.py
+++ b/broker/api.py
@@ -119,9 +119,8 @@ class API(ServiceBroker):
 
         params = details.parameters or {}
 
-        if params.get("domains"):
-            domain_names = [d.strip().lower() for d in params["domains"].split(",")]
-        else:
+        domain_names = parse_domain_options(params)
+        if not domain_names:
             raise errors.ErrBadRequest("'domains' parameter required.")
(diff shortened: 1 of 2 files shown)


broker/api.py
```diff
@@ -386,3 +383,11 @@ def parse_header_options(params):
...
     headers = {header.upper() for header in headers}
     return sorted(list(headers))
+
+
+def parse_domain_options(params):
+    domains = params.get("domains", None)
+    if isinstance(domains, str):
+        return [d.strip().lower() for d in domains.split(",")]
+    if isinstance(domains, list):
+        return domains
```
apburnes, line 393: Should we clean the strings when it's a list too?

bengerman13: that would probably make sense.

bengerman13: fixed in bc98c399731e720b1e94faec7a7cbc3f80bcf012

apburnes (approved):
💯

PR #184, opened 2020-10-20
add MigrationServiceInstance class

bengerman13:

- Add a MigrationServiceInstance class and a provisioning path for it
- refactor CDN provisioning a tiny bit to make the switch statement on plans a little easier to follow

## Security considerations

None

### DIFF
--- a/tests/integration/migration/test_migration_provisioning.py
+++ b/tests/integration/migration/test_migration_provisioning.py
@@ -1,0 +1,106 @@
+import json
+from datetime import date, datetime
+
+import pytest  # noqa F401
+
+from broker.extensions import config, db
+from broker.models import Challenge, Operation, CDNServiceInstance
+from tests.lib.factories import CDNServiceInstanceFactory
+from tests.lib.client import check_last_operation_description
+
+from tests.integration.cdn.test_cdn_update import (
+    subtest_update_happy_path,
+    subtest_update_same_domains,
+)
+
+# The subtests below are "interesting".  Before test_provision_happy_path, we
+# had separate tests for each stage in the task pipeline.  But each test would
+# have to duplicate much of the previous test.  This was arduous and slow. Now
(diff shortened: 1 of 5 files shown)


tests/lib/client.py
```diff
@@ -78,6 +78,26 @@ def provision_cdn_instance(
...
+            "plan_id": "739e78F5-a919-46ef-9193-1293cc086c17",
+            "organization_guid": "abc",
+            "space_guid": "123",
+        }
+
+        if params is not None:
+            json["parameters"] = params
+
+        self.put(
+            f"/v2/service_instances/{id}",
```
ccostino, line 95: Should this be a `/v3` API call now, or is that not quite ready with this?

ccostino (commented):
Thanks, @bengerman13!  I left a note for others taking a look regarding the `noqa` comments, and only had a question regarding an API call.

tests/lib/client.py
bengerman13: this is the service broker, not CAPI, so `v2` is current :)

tests/integration/migration/test_migration_provisioning.py
```diff
@@ -0,0 +1,106 @@
+import json
+from datetime import date, datetime
+
+import pytest  # noqa F401
```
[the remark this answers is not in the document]
bengerman13: It's actually also irrelevant here - we stopped using the linter and use `black` instead now - it's here because of copy-paste

The reason we need this for the linter is that we're importing pytest but not using it, though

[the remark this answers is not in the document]
ccostino: Ah good to know!  Thanks. :-)

tests/lib/client.py
ccostino: Thanks for the clarification! :-)

ccostino (approved):
Thanks, @bengerman13!

PR #185, opened 2020-10-22
add framework for updating service instance type

bengerman13:
This is a little clumsy because, as far as I can tell, sqlalchemy won't
let us just directly convert a model type.

- add framework for updating service instance type

## Security considerations

None

### DIFF
--- a/broker/api.py
+++ b/broker/api.py
@@ -76,10 +76,10 @@ class API(ServiceBroker):
                     description="Custom domain with TLS and CloudFront.",
                 ),
                 ServicePlan(
-                    id=MIGRATION_PLAN_ID, 
+                    id=MIGRATION_PLAN_ID,
                     name="migration-not-for-direct-use",
                     description="Migration plan for internal autmation.",
-                )
+                ),
             ],
         )
 
@@ -142,13 +142,14 @@ class API(ServiceBroker):
             instance = ALBServiceInstance(id=instance_id, domain_names=domain_names)
             queue = queue_all_alb_provision_tasks_for_operation
         elif details.plan_id == MIGRATION_PLAN_ID:
-            instance = MigrationServiceInstance(id=instance_id, domain_names=domain_names)
+            instance = MigrationServiceInstance(
+                id=instance_id, domain_names=domain_names
+            )
             db.session.add(instance)
             db.session.commit()
             return ProvisionedServiceSpec(state=ProvisionState.SUCCESSFUL_CREATED)
         else:
             raise NotImplementedError()
-
 
         self.logger.info("setting origin hostname")
         self.logger.info("creating operation")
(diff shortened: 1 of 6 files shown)


broker/api.py
```diff
@@ -76,10 +76,10 @@ def catalog(self) -> Service:
                     description="Custom domain with TLS and CloudFront.",
                 ),
                 ServicePlan(
-                    id=MIGRATION_PLAN_ID, 
+                    id=MIGRATION_PLAN_ID,
                     name="migration-not-for-direct-use",
                     description="Migration plan for internal autmation.",
-                )
+                ),
```
bengerman13, line 82: These are just from running `black` - I forgot to run it last time I pushed 😬

broker/api.py
```diff
@@ -142,14 +142,15 @@ def provision(
             instance = ALBServiceInstance(id=instance_id, domain_names=domain_names)
             queue = queue_all_alb_provision_tasks_for_operation
         elif details.plan_id == MIGRATION_PLAN_ID:
-            instance = MigrationServiceInstance(id=instance_id, domain_names=domain_names)
+            instance = MigrationServiceInstance(
+                id=instance_id, domain_names=domain_names
+            )
```
bengerman13, line 147: also just formatting from `black`

broker/api.py
```diff
@@ -401,11 +403,7 @@ def provision_cdn_instance(instance_id: str, domain_names: list, params: dict):
...
             )
-        instance.origin_protocol_policy = (
-            CDNServiceInstance.ProtocolPolicy.HTTP.value
-        )
+        instance.origin_protocol_policy = CDNServiceInstance.ProtocolPolicy.HTTP.value
     else:
-        instance.origin_protocol_policy = (
-            CDNServiceInstance.ProtocolPolicy.HTTPS.value
-        )
+        instance.origin_protocol_policy = CDNServiceInstance.ProtocolPolicy.HTTPS.value
```
bengerman13, line 408: More stuff from `black`

broker/api.py
apburnes: Should we add a lint step to CI?

bengerman13: 🤷 I'm not that concerned about making sure stuff stays 100% linted - I just use `black` periodically so it doesn't get out of control. I'd hate to have a necessary change have to go through a second round of PRs because of an extra space
since `black` is a formatter, we can't add it to CI unless/until we figure out a way to let bots commit

apburnes (approved):

broker/models.py
```diff
@@ -213,3 +213,48 @@ class Challenge(Base):
...
+        MigrationServiceInstance,
+    ):
+        # we only know about service instances
+        raise NotImplementedError()
+    if not isinstance(service_instance, ServiceInstance):
+        # we only know about service instances
+        raise NotImplementedError()
+    if type(service_instance) == new_type:
+        # why are you doing this?
+        return service_instance
```
ccostino, line 248: Why indeed...  I feel like this is the path of least resistance though as opposed to raising an error and going that route, so I approve. :-)

broker/api.py
apburnes: Ah, just realized we don't have CI setup on PR's. Does caulking not like other pre-commit hooks on repos?

ccostino (approved):
This looks good to me, thanks @bengerman13!

broker/api.py
bengerman13: It does not: cloud-gov/caulking#37

PR #186, opened 2020-10-24
Update migration instances

bengerman13:

- Add the rest of the migration-to-cdn update path

## Security considerations

None


broker/api.py
```diff
@@ -293,8 +302,22 @@ def update(
...
             queue = queue_all_alb_update_tasks_for_operation
+        elif instance.instance_type == "migration_service_instance":
+            if details.plan_id == CDN_PLAN_ID:
+                noop = False
+                validate_migration_to_cdn_params(params)
+                instance = change_instance_type(
+                    instance, CDNServiceInstance, db.session
+                )
+                update_cdn_params_for_migration(instance, params)
+                db.session.add(instance.current_certificate)
```
ccostino, line 317: Is the current certificate being added explicitly here only in the case of migrations because it's needed during that time?

Furthermore, `update_cdn_params_for_migrations` sets a bunch of other things based on the `params` passed in if this code path is reached, but does anything actually happen with those values?

broker/api.py
```diff
@@ -407,3 +430,60 @@ def provision_cdn_instance(instance_id: str, domain_names: list, params: dict):
...
+    instance.current_certificate = Certificate(service_instance_id=instance.id)
+    instance.current_certificate.iam_server_certificate_id = params[
+        "iam_server_certificate_id"
+    ]
+    instance.current_certificate.iam_server_certificate_arn = params[
+        "iam_server_certificate_arn"
+    ]
+    instance.current_certificate.iam_server_certificate_name = params[
+        "iam_server_certificate_name"
+    ]
```
ccostino, line 481: Just wanted to check to make sure nothing else had to happen here at the end after setting all of these things!

broker/app.py
```diff
@@ -43,12 +43,19 @@ def ping():
...
+    @app.errorhandler(ClientError)
+    def handle_not_implemented(e):
         logger.exception(e)
         return (
-            to_json_response(
-                ErrorResponse(description="Unhandled error during request")
-            ),
-            HTTPStatus.INTERNAL_SERVER_ERROR,
+            to_json_response(ErrorResponse(description=str(e))),
+            HTTPStatus.NOT_IMPLEMENTED,
```
ccostino, line 58: Thanks for making this clearer to help with debugging!

ccostino (commented):
Thanks, @bengerman13!  I had a couple of questions and definitely fair on the single test files for future expansion, it still helps keep things organized and easier to reason about. :-)

broker/api.py
bengerman13: nope - because this is an object passed by reference, this is just about updating its params. The only reason I pulled it out into a function was so the if-elif-else chain above would read more easily

ccostino: Makes sense!

ccostino (approved):
Thanks, @bengerman13!

PR #196, opened 2021-03-17
allow arbitrary-depth of CNAMEs

### LINKED ISSUE #195: Recurse CNAMEs for validation
We currently support only one level of CNAME when validating _acme-challenge names. We should support arbitrary depths of CNAME indirection.

Change needs to happen here:
https://github.com/cloud-gov/external-domain-broker/blob/1f7a890c24539147dbbf7f2528805fd5af12665e/broker/validators.py#L32

from the report from data.gov:

```
cf create-service external-domain domain test-cloud-gov-indirect-api-data-gov-ext-domain -c '{"domains": "test-cloud-gov-indirect.api.data.gov"}'
Service broker error: We could not find correct CNAME records for one or more of your domains.
Please ensure the following DNS records are in place and try to provision
this service again:
  CNAME _acme-challenge.test-cloud-gov-indirect.api.data.gov should point to _acme-challenge.test-cloud-gov-indirect.api.data.gov.external-domains-production.cloud.gov, but it is set incorrectly to _acme-challenge.test-cloud-gov-indirect-value.api.data.gov.
```

>In this case, _acme-challenge.test-cloud-gov-indirect.api.data.gov is a CNAME that points to _acme-challenge.test-cloud-gov-indirect-value.api.data.gov which is then a CNAME pointing to the expected value of _acme-challenge.test-cloud-gov-indirect.api.data.gov.external-domains-production.cloud.gov. However, Cloud.gov only looks at the first CNAME, so it doesn’t consider this valid. We’re basically wondering about the feasibility of making something like this setup valid.

bengerman13:
fixes #195

- allow arbitrary-depth of CNAMEs

## Security considerations

None

### DIFF
--- a/tests/integration/cdn/test_cdn_update.py
+++ b/tests/integration/cdn/test_cdn_update.py
@@ -815,7 +815,4 @@ def subtest_update_updates_ALIAS_records(tasks, route53):
     route53.assert_no_pending_responses()
     db.session.expunge_all()
     service_instance = CDNServiceInstance.query.get("4321")
-    assert service_instance.route53_change_ids == [
-        bar_com_change_id,
-        foo_com_change_id,
-    ]
+    assert service_instance.route53_change_ids == [bar_com_change_id, foo_com_change_id]
--- a/broker/validators.py
+++ b/broker/validators.py
@@ -31,19 +31,28 @@ class CNAME:
     def _error_for_domain(self, domain: str) -> str:
         cname = get_cname(acme_challenge_cname_name(domain))
 
+        # track the CNAMEs we've resolved to check for loops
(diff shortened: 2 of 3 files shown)


tests/integration/cdn/test_cdn_update.py
```diff
@@ -815,7 +815,4 @@ def subtest_update_updates_ALIAS_records(tasks, route53):
     route53.assert_no_pending_responses()
     db.session.expunge_all()
     service_instance = CDNServiceInstance.query.get("4321")
-    assert service_instance.route53_change_ids == [
```
bengerman13, line 818: This change is unrelated. Looks like someone forgot to run `black` previously

onelittlebecca (approved):

PR #198, opened 2021-04-14
bump jinja (and others) to fix Jinja vuln

bengerman13:

- This bumps Jinja to fix a vuln. We're not actually using jinja, so the vuln isn't applicable to us, but dependabot was bugging me about it

pip-tools changed its formatting, too, which makes it look like everything got bumped, but most deps didn't change.

## Security considerations

Fixes a theoretical vuln, but doesn't actually apply to us


ccostino (approved):
Thanks, @bengerman13!

PR #215, opened 2021-07-23
clean up new_certificate when retrieval fails due to validation errors

bengerman13:

- clean up new_certificate when retrieval fails due to validation errors. When a validation fails, we need a new order and new challenges in order to retry. Because of this, we need to delete the service_instance.new_certificate record, so we don't try to reuse it

## Security considerations

None

### DIFF
--- a/broker/tasks/letsencrypt.py
+++ b/broker/tasks/letsencrypt.py
@@ -238,7 +238,14 @@ def answer_challenges(operation_id: int, **kwargs):
         )
         challenge_response = challenge_body.response(wrapped_account_key)
         # Let the CA server know that we are ready for the challenge.
-        client_acme.answer_challenge(challenge_body, challenge_response)
+        response = client_acme.answer_challenge(challenge_body, challenge_response)
+        print(response)
+        if response.body.error is not None:
+            # log the error for now. We haven't reproduced this locally, so we can't act on it yet
+            # but it would be interesting in the real world
+            logger.error(
+                f"challenge for instance {service_instance.id} errored. Error: {response.body.error}"
+            )
         challenge.answered = True
         db.session.add(challenge)
         db.session.commit()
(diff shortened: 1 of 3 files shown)


broker/tasks/letsencrypt.py
```diff
@@ -329,11 +336,18 @@ def cert_from_fullchain(fullchain_pem: str) -> str:
...
-                f"failed to retrieve certificate for {service_instance.domain_names} with errors {e.failed_authzrs}"
-            )
-            raise e
-
+        logger.error(
+            f"failed to retrieve certificate for {service_instance.domain_names} with errors {e.failed_authzrs}"
+        )
+        # if we fail validation, nuke the cert record and its challenges.
+        # this way, when we retry from the beginning, we won't try to reuse them
+        # the bad new is that we'll still retry this task a bunch of times before the pipeline fails
```
ccostino, line 344: The cert will still fail in future retries after eliminating the remnants from the failed attempt at renewing?  Do we just have to let it run its course, or is there something else that can be done?

ccostino (commented):
Thanks, @bengerman13!  One quick question from me.

onelittlebecca (approved):
dang dude! well done!

broker/tasks/letsencrypt.py
bengerman13: right - when it fails once, we nuke all this stuff, then on all the other retries it will continue failing (but differently - probably running into `Nonetype doesn't have attribute ...` ). We have a fixed limit on retries, so eventually we'll fail the whole pipeline. On a provision or update operation, this will cause the broker & CAPI to see this as a failure (which would've happened anyway).
Renewals are where this really matters. By cleaning these resources up, the next time cron triggers a new renewal pipeline for this instance, it'll get a fresh start. This means (unlike the existing case) if the error is intermittent (e.g. dns error, race condition, etc), we have hope of renewing the next time around.

ccostino: Thank you for explaining!

PR #231, opened 2022-01-15
ignore duplicate domains in special environments

bengerman13:

- ignore duplicate domains in special environments

## Security considerations

None


ccostino (approved):
Thanks, @bengerman13!

PR #248, opened 2022-12-21
Add job to create metrics for duplicate service certificates

markdboyd:

- Add CI job to create metrics for duplicate service certificates
- Add Docker files to facilitate running PG database for local testing/debugging
- Add code to find brokered domain services with multiple certificates in the database

## Security considerations

There shouldn't be any changes to security. We are just adding new metrics for the purposes of alerting and observability


docker/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
```
markdboyd, line 1: I created this docker-compose set up so that I could easily debug tests that interact with the database. 

The existing setup by default ran all the tests from **inside a Docker container**, which leaves no way to attach a debugger. This setup just runs PG as a service exposed on port 5432 of the host, which is what the configuration for local testing expects

docker/postgresql/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+  && echo "log_connections = on" >> "$PGCONFIG" \
+  && echo "log_disconnections = on" >> "$PGCONFIG" \
+  && echo "log_line_prefix = ''" >> "$PGCONFIG" \
+  && sed -ri "s!^#?(listen_addresses)\s*=\s*\S+.*!\1 = '*'!" "$PGCONFIG" \
+  && grep -F "listen_addresses = '*'" "$PGCONFIG" \
+  && echo 'CREATE DATABASE "test"' | postgres --single -D "$PGDATA" postgres \
+  && echo 'CREATE DATABASE "local-development"' | postgres --single -D "$PGDATA" postgres
+
+# Trust connection from Docker for local development
+RUN echo "host local-development ${USER} 172.16.0.0/12 trust" >> /tmp/data/pg_hba.conf
```
markdboyd, line 44: I'm very unsure about this. 

It's necessary to add something like this so that the testing database in the container can be accessed from the host machine. See https://www.postgresql.org/docs/current/auth-pg-hba-conf.html.

This Dockerfile is only used for local testing, so it's not like we're opening a security hole into an actual public-facing service.

Also, 172.16.0.0/12 is a reserved IP for local networking: https://en.wikipedia.org/wiki/Reserved_IP_addresses, so the idea is to make sure we're only allowing ingress from local IP ranges.

So it seems pretty safe, but it kinda freaks me out still

rcgottlieb (approved):

jameshochadel (approved):

PR #249, opened 2022-12-22
Add code and pipeline job to duplicate certs

markdboyd:

- Add pipeline job to remove duplicate certs
- Add code to remove duplicate certificate records from database and delete associated AWS certificate resources
- Add unit tests for code

## Security considerations

There shouldn't be any security impact


docker/postgresql/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM postgres:11
 ARG UID=2000
 ARG GID=2000
 ARG USER=pg-user
+ARG POSTGRES_PASSWORD
```
jameshochadel, line 6: I notice you're using some of the postgres-supported `PG` variables below; what do you think about replacing this with `PGPASSWORD`, which (I think) will be picked up automatically by the postgres CLI and library?

markdboyd: I have no objection to changing it, but the key thing is that the password needs to be used by `initdb` when creating the database:

https://github.com/cloud-gov/external-domain-broker/pull/249/files#diff-ad94bb83eca169b86d41ed814eb5b046d45daf895147df1d152ef4949257ffabR37

I'm not sure if `initdb` respects `PGPASSWORD` by default, but I can try it

markdboyd: Yeah, as I suspected, you still have to manually pass the value into `initdb`, but there's no harm in changing to that environment variable anyway

8a2311a2b69b1e16d2cc3bc925b6dd6648e3fcfc

jameshochadel: Ah, that's my bad, I shouldn't have assumed. Ok, thanks regardless.

jameshochadel (approved):

PR #262, opened 2023-01-20
Update metric name and certificate removal logic

markdboyd:

- Update metric name for duplicate certs to `service_instance_duplicate_cert_count` to clarify that it is a count of duplicate certificates for a service, not all certificates for a service. There should be a total of `service_instance_duplicate_cert_count` + 1 (current certificate) certificates for a service
- Update logic for removing certificates from load balancer listeners to wait for the certificate to no longer be listed in the certificates for that listener. This fixes a bug I observed in the removal where attempting to delete the certificate itself immediately after removing it from the listener failed because the removal from the listener hadn't fully propagated through AWS

## Security considerations

None

### DIFF
--- a/broker/duplicate_certs.py
+++ b/broker/duplicate_certs.py
@@ -64,15 +64,20 @@ def get_listener_cert_arns(listener_arn, alb=alb):
     listener_cert_arns = [cert["CertificateArn"] for cert in response["Certificates"]]
     return listener_cert_arns
 
-def remove_certificate_from_listener_and_wait_for_deletion(listener_arn, certificate_arn):
+def remove_certificate_from_listener_and_verify_removal(listener_arn, certificate_arn, alb=alb):
     alb.remove_listener_certificates(
         ListenerArn=listener_arn,
         Certificates=[{"CertificateArn": certificate_arn}]
     )
+    attempts = 0
     while True:
-        listener_cert_arns = get_listener_cert_arns(listener_arn)
+        if (attempts == 10):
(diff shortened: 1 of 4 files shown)


broker/duplicate_certs.py
```diff
@@ -57,16 +57,35 @@ def delete_iam_server_certificate(certificate_name):
...
+    attempts = 0
+    while True:
+        if (attempts == 10):
+            logger.info(f"Could not verify certificate {certificate_arn} was removed from listener {listener_arn} after 10 tries, giving up")
+            break
+        listener_cert_arns = get_listener_cert_arns(listener_arn, alb=alb)
+        if certificate_arn not in listener_cert_arns:
+            break
+        attempts += 1
+    logger.info(f"Removed certificate {certificate_arn} from listener {listener_arn}")
```
jameshochadel, line 81: If I may suggest a simplification:

```suggestion
    for _ in range(10):
        listener_cert_arns = get_listener_cert_arns(listener_arn, alb=alb)
        if certificate_arn not in listener_cert_arns:
            logger.info(f"Removed certificate {certificate_arn} from listener {listener_arn}")
            return
    logger.info(f"Could not verify certificate {certificate_arn} was removed from listener {listener_arn} after 10 tries, giving up")
```

Also, did you consider using a different log level for the failure case? Not sure if it provides any value, but thought I'd ask.

jameshochadel (commented):
One small suggestion, otherwise LGTM.

jameshochadel (dismissed):

jameshochadel (approved):

Issue #283, opened 2024-02-05
Update boto3 version to currnet


ChrisMcGowan:
AWS is alerting cloud.gov that we are using deprecated versions of boto3 health SDK and need to above versions `1.27.1`  to work.  As of Feb'24 version `1.34.34` is the most recent version.

## Notes

- Validate that recent versions don't introduce new issues/bugs

## Acceptance Criteria

- [ ] Determine code changes needed for the upgrade
- [ ] Throughly test in lower environments
- [ ] If a test suite does not exist in the repo - create on to validate new versions of `boto3` in the future

markdboyd:
This change has been deployed to production and successfully passed the acceptance tests there

PR #288, opened 2024-02-12
Update boto3 version

markdboyd:

After #285 was merged, the tests started failing for this app. And it quickly got very complicated to try and update all the dependencies to their latest versions and get the tests to pass.

So I reverted most dependencies to their versions as of 43ebfaf8caa9122cdb8942a579e2b5d16e5a8f57 and then did a `pip-tools compile --upgrade` to ensure that `boto3` was upgraded. I had to add version constraints for several dependencies to `pip-tools/requirements.in` to avoid deprecation warnings and other version mismatch issues, since the primary goal of this ticket, #283, is simply to upgrade`boto3`.

The tests are now passing with these changes.

The PR also updates the pipeline to include notifications when the testing for this broker fails.

## Things to check

## Security considerations

Some dependencies are being upgraded to later versions, which is good for security and general project hygiene


requirements.txt
```diff
@@ -126,23 +126,24 @@ six==1.16.0
...
-    #   alembic
-    #   sqlalchemy
+    # via alembic
 urllib3==1.26.18
     # via
     #   botocore
     #   requests
-werkzeug==3.0.1
-    # via flask
+werkzeug==2.0.3
```
rcgottlieb, line 143: Not sure if this impacts this repo, but the changes made to cf-hello-worlds modified two packages. one of which is Werkzeug 2.2.2:
https://github.com/cloud-gov/cf-hello-worlds/pull/127/files

rcgottlieb: Ah good to know.

rcgottlieb (approved):

PR #293, opened 2024-03-20
Add provisioning, update, and renewal for dedicated alb instances

bengerman13:

- Add new dedicated alb instance type with provisioning, update, and renewal pipelines
- improve tests script to operate faster and work in cases where ionotify is not available
- tidy README
- refactor shared/copy-paste tests into parametrized tests
- blacken everything (sorry - this introduced a lot of noise because the code wasn't blackened before I started😬)

### High-level view

This _starts_ adding the `domain-with-org-lb` plan (remaining work described below).
This plan allows specific ALBs to tie to specific orgs. In the current iteration, this works by specifying a second list of ALBs to the broker, which are available as DedicatedALBs. 
When a user requests a `domain-with-org-lb` instance, the broker checks if there are DedicatedALBs associated with their org. If there are none, it then checks if there are any DedicatedALBs associated to _no_ orgs. If there are any, it grabs a random one and associates it to the user's org.

### Remaining work

We still need:
- a way to move between a `domain` plan and a `domain-with-org-lb` plan
- a deprovision path for `domain-with-org-lb` instances
- detection for when all DedicatedALBs within an org are full or nearing full, so we can pick a new DedicatedALB for the org
- alerting for too few available DedicatedALBs
- alerting for too few available certificates in an org
- ci plumbing for listing DedicatedALBs

We think we also want:
- a way to declare in config that certain ALBs are only for specific instances, before any org gets a shot at them. This will allow us to provision ALBs that point to dedicated gorouters for specific customer needs
- a way to share back a DedicatedALB after the last instance from a org abandons it

## Things to check

## Security considerations

None


tests/integration/alb/test_alb_remove_duplicate_certs.py
bengerman13, line 1: look out - GitHub thinks there's no way this file matters, but it does

tests/integration/dedicated_alb/test_dedicated_alb_update.py
bengerman13, line 1: Look out - GitHub thinks this file is too big to be real, but it's real

broker/models.py
```diff
@@ -150,19 +150,37 @@ def __repr__(self):
...
         return f"<ALBServiceInstance {self.id} {self.domain_names}>"
 
 
+class DedicatedALBServiceInstance(AbstractALBServiceInstance):
+    org_id = db.Column(db.String)
+
+    __mapper_args__ = {"polymorphic_identity": "dedicated_alb_service_instance"}
+
+    def __repr__(self):
+        return f"<ALBServiceInstance {self.id} {self.domain_names}>"
```
markdboyd, line 181:
```suggestion
        return f"<DedicatedALBServiceInstance {self.id} {self.domain_names}>"
```

tests/integration/alb/test_alb_provisioning.py
```diff
@@ -69,103 +68,6 @@ def test_gets_lowest_used_alb(alb):
     ) == ("alb-listener-arn-1", "listener-arn-1")
 
 
-def test_refuses_to_provision_synchronously(client):
```
markdboyd, line 72: Why are we deleting these tests?

tests/integration/cdn/test_cdn_provisioning.py
```diff
@@ -22,104 +22,6 @@
 # these subtasks when testing failure scenarios.
 
 
-def test_refuses_to_provision_synchronously(client):
```
markdboyd, line 25: Again, why are we removing these tests?

tests/integration/dedicated_alb/test_dedicated_alb_provisioning.py
```diff
@@ -0,0 +1,332 @@
...
+# The subtests below are "interesting".  Before test_provision_happy_path, we
+# had separate tests for each stage in the task pipeline.  But each test would
+# have to duplicate much of the previous test.  This was arduous and slow. Now
+# we have a single test_provision_happy_path, and many subtest_foo helper
+# methods.  This still makes it clear which stage in the task pipeline is
+# failing (look for the subtask_foo in the traceback), and allows us to re-use
+# these subtasks when testing failure scenarios.
+
+
+def test_provision_happy_path(
```
markdboyd, line 32: I think my main question is: why can't we test by invoking the huey pipeline for provisioning?

tests/unit/test_config.py
```diff
@@ -78,10 +78,15 @@ def mocked_env(monkeypatch, vcap_application, vcap_services):
...
     monkeypatch.setenv("VCAP_SERVICES", vcap_services)
     monkeypatch.setenv("DEFAULT_CLOUDFRONT_ORIGIN", "None")
+    # TODO: validate listener arns are unique
     monkeypatch.setenv(
         "ALB_LISTENER_ARNS",
         "arn:aws:elasticloadbalancing:us-east-2:123456789012:listener/app/my-load-balancer/1234567890123456,arn:aws:elasticloadbalancing:us-east-2:123456789012:listener/app/my-load-balancer/1234567890123456",
     )
+    monkeypatch.setenv(
+        "DEDICATED_ALB_LISTENER_ARNS",
+        "arn:aws:elasticloadbalancing:us-east-2:123456789012:listener/app/my-other-load-balancer/1234567890123456,arn:aws:elasticloadbalancing:us-east-2:123456789012:listener/app/my-other-load-balancer/1234567890123456",
```
markdboyd, line 88: Are we intentionally setting the value to the same exact listener ARN twice so that we can verify that de-duping works properly and only unique values are loaded into the config?

broker/tasks/alb.py
```diff
@@ -25,6 +31,68 @@ def get_lowest_used_alb(listener_arns):
...
+def load_albs(dedicated_listener_arns):
+    stmt = insert(DedicatedALBListener).values(
+        [dict(listener_arn=arn) for arn in dedicated_listener_arns]
+    )
+    stmt = stmt.on_conflict_do_nothing(index_elements=["listener_arn"])
+    db.session.execute(stmt)
+
+
+def get_lowest_dedicated_alb(service_instance, db):
+    # TODO: handle grabbing the next available ALB when ours are full, or nearly so
```
markdboyd, line 44: How are the TODOs going to be handled? Follow up PRs?

broker/tasks/alb.py
```diff
@@ -25,6 +31,68 @@ def get_lowest_used_alb(listener_arns):
...
+    operation.step_description = "Selecting load balancer"
+    flag_modified(operation, "step_description")
+    db.session.add(operation)
+    db.session.commit()
+
+    if (
+        service_instance.alb_arn
+        and operation.action == Operation.Actions.PROVISION.value
+    ):
+        return
```
markdboyd, line 90: Why do we just return on provision? I thought this is where we would need to select the lowest dedicated ALB for the domain?

tests/integration/migration/test_migration_provisioning.py
```diff
@@ -1,108 +0,0 @@
-import json
```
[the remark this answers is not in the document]
markdboyd: Moved to `common_checks.py`? But I don't see any tests invoking `client.provision_migration_instance` in `common_checks.py`?

tests/integration/alb/test_alb_provisioning.py
bengerman13: these tests were identical except for the plan type we were calling, so I merged them all into `common_checks.py`, where we test all of them using `@pytest.mark.parametrize` (except I missed migration instances, as you pointed out in another comment - I'll fix that)

tests/unit/test_config.py
bengerman13: this config is only used for testing that the config gets loaded, so it doesn't validate any de-dupe and isn't used outside this file.
the dedicated_alb_listener_arns is duplicated because I copied the alb_listener_arns and tweaked it slightly. I assume the alb_listener_arns is duplicated because I was just as lazy when I wrote that part

broker/tasks/alb.py
bengerman13: yeah - that's the plan (although I'm working on that today, so it might end up just ballooning this PR instead)

markdboyd: I might prefer separate follow up PRs if possible. It's much easier to review the code in more digestible chunks

tests/integration/dedicated_alb/test_dedicated_alb_provisioning.py
markdboyd: To be more precise, why can't we use Huey itself to run the pipeline in the test?

broker/tasks/alb.py
bengerman13: this is an idempotency catch - if we're provisioning and we already have an alb_arn, we already did this step and can jump along to the next one. 

These types of checks are for when we re-enqueue a pipeline that got orphaned

tests/integration/dedicated_alb/test_dedicated_alb_provisioning.py
bengerman13: under the hood, we are using huey to run the pipeline, but we're forcing it to only grab one task at a time (we do this [here](https://github.com/cloud-gov/external-domain-broker/blob/cc05cf3341dc7e774e97257c950d4db78ed36a9a/tests/lib/tasks.py#L52-L76))
if you're asking why we don't just hand Huey the whole pipeline and check on it at the end:
- I think it's a lot easier to track our mocks/fixtures/etc if we handle them one step at a time; if we did the whole pipeline, we'd have to queue up every expected AWS call up front 
- having the subtests lets us test idempotency of each task as we go
- it's easier to assure that we're doing the right things and in our correct state all the way down

tests/integration/dedicated_alb/test_dedicated_alb_pick_listener.py
```diff
@@ -0,0 +1,92 @@
...
+            other_listener_1,
+        ]
+    )
+    clean_db.session.commit()
+
+    clean_db.session.expunge_all()
+    service_instance = DedicatedALBServiceInstance.query.get("4321")
+    alb.expect_get_certificates_for_listener("our-arn-0", 1)
+    alb.expect_get_certificates_for_listener("our-arn-1", 5)
+    alb.expect_get_listeners("our-arn-0")
```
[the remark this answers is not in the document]
bengerman13: `FakeALB.expect_get_certificates_for_listener` corresponds to `ALB.describe_listener_certificates`
`FakeALB.expect_get_listeners` corresponds to `ALB.describe_listeners`

The task runs `ALB.describe_listener_certificates` against each potential listener to see which has the fewest certificates attached. 
Once that's done, it knows which listener to use. The call to `ALB.describe_listeners` is to get more metadata about the chosen listener, so we only expect it to be run against `"our-arn-0"`
(specifically, it's looking to get the listener's alb arn)

tests/integration/migration/test_migration_provisioning.py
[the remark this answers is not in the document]
bengerman13: yeah, exactly. Good catch that I missed `migration`s though

tests/unit/test_config.py
```diff
@@ -189,6 +195,18 @@ def test_config_provides_alb_arns(env, monkeypatch, mocked_env):
...
 
+@pytest.mark.parametrize("env", ["production", "staging", "development"])
+def test_config_provides_dedicated_alb_arns(env, monkeypatch, mocked_env):
+    monkeypatch.setenv("FLASK_ENV", env)
+
+    config = config_from_env()
+
+    assert type(config.DEDICATED_ALB_LISTENER_ARNS) == list
+    assert config.DEDICATED_ALB_LISTENER_ARNS == [
+        "arn:aws:elasticloadbalancing:us-east-2:123456789012:listener/app/my-other-load-balancer/1234567890123456"
```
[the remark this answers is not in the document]
bengerman13: 🤦🏻 I misspoke when I said there's no deduplication. We dedupe [here](https://github.com/cloud-gov/external-domain-broker/blob/cc05cf3341dc7e774e97257c950d4db78ed36a9a/broker/config.py#L58)

tests/integration/migration/test_migration_provisioning.py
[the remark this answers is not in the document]
bengerman13: wait, I didn't miss migration 😅 
the new tests all leverage [`CFAPIClient.provision_instance`](https://github.com/cloud-gov/external-domain-broker/blob/d5db6c230c6d33f2f3e84a17f406fb7748e5f06f/tests/lib/client.py#L62-L71), which takes the instance type as a string and calls the appropriate provision function

tests/unit/test_config.py
bengerman13: ☝🏻 I was wrong about deduplication - we are deduplicating and we are testing it here. 
I added comments to the fixture and tests about that.

tests/unit/test_config.py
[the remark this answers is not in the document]
bengerman13: added comments about dedupe and its testing

markdboyd (approved):

PR #325, opened 2024-04-18
Bb/better dedicated alb sourcing

bengerman13:

- dedicate a new ALB to an org when too few slots are left on the existing ALB
- fix some comments
- use `sqlalchemy.null` instead of `None` so we can shake some `# noqa`s

## Things to check

## Security considerations

None

### DIFF
--- a/broker/config.py
+++ b/broker/config.py
@@ -54,8 +54,8 @@ class Config:
         # https://docs.aws.amazon.com/Route53/latest/APIReference/API_AliasTarget.html
         self.CLOUDFRONT_HOSTED_ZONE_ID = "Z2FDTNDATAQYW2"
 
-        # the maximum from AWS is 20, and we alert when we have 18 on a given alb
-        self.MAX_CERTS_PER_ALB = 17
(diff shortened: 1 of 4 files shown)


broker/config.py
```diff
@@ -54,6 +54,9 @@ def __init__(self):
         # https://docs.aws.amazon.com/Route53/latest/APIReference/API_AliasTarget.html
         self.CLOUDFRONT_HOSTED_ZONE_ID = "Z2FDTNDATAQYW2"
 
+        # the maximum from AWS is 20, and we alert when we have 18 on a given alb
+        self.MAX_CERTS_PER_ALB = 17
```
markdboyd, line 58: I thought it was 25?

> Certificates per Application Load Balancer (excluding default certificates)	25

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html

bengerman13: good catch - updated the comment and value

markdboyd (approved):

PR #337, opened 2024-06-28
Update CDN dedicated WAF plan provision logic

markdboyd:

- Add pipeline tasks for:
  - Creating a custom web ACL
  - Creating Route53 health check(s) for domain(s)
  - Associated Route53 health check(s) with the Shield Advanced protection. [Having Route53 health checks associated with Shield Advanced protection enables proactive engagement from AWS Shield Response team](https://docs.aws.amazon.com/waf/latest/developerguide/ddos-srt-proactive-engagement.html).
- Add/update tests
- Add `dev` command for running containers with `docker-compose`

## Things to check

## Security considerations

There should be no direct security considerations to the code changes. But the effect of changes should be that customers can create CDNs with dedicated WAF ACLs, which allows better customization of WAF rules to their traffic patterns and more effective mitigation by Shield Advanced


broker/tasks/pipelines.py
```diff
@@ -136,11 +137,13 @@ def queue_all_cdn_dedicated_waf_provision_tasks_for_operation(
...
         .then(iam.upload_server_certificate, operation_id, **correlation)
-        .then(cloudfront.create_distribution, operation_id, **correlation)
+        .then(waf.create_web_acl, operation_id, **correlation)
+        .then(cloudfront.create_distribution_with_tags, operation_id, **correlation)
         .then(cloudfront.wait_for_distribution, operation_id, **correlation)
         .then(route53.create_ALIAS_records, operation_id, **correlation)
         .then(route53.wait_for_changes, operation_id, **correlation)
-        .then(waf.create_web_acl, operation_id, **correlation)
+        .then(route53.create_health_checks, operation_id, **correlation)
+        .then(shield.associate_health_checks, operation_id, **correlation)
```
markdboyd, line 146: This might not work because the Shield protection is added automatically by AWS and might not be ready by the time this step is reached, but it seemed worth trying because other approaches are much more complex

broker/lib/tags.py
```diff
@@ -0,0 +1,11 @@
+def add_tag(tags, tag_key: str, tag_value: str):
+    if not tags:
+        tags = {}
+    tags["Items"] = tags["Items"] if "Items" in tags else []
```
bengerman13, line 4: this could be:
```suggestion
    tags["Items"] = tags.get("Items", [])
```

bengerman13 (dismissed):

bengerman13 (approved):

broker/tasks/shield.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_cloudfront_protections(self):
+        if not self.protected_cloudfront_ids:
+            self._list_cloudfront_protections()
+        return self.protected_cloudfront_ids
+
+
+# Initialize here so that the class and its dictionary of mapped protections are
+# kept in memory for use by tasks
+shield_protections = ShieldProtections()
```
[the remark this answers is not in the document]
markdboyd: I couldn't do this because it creates a circular dependency. `ShieldProtections` requires `shield` from `broker/aws.py` which itself imports `broker/extensions.py`

bengerman13 (approved):

PR #338, opened 2024-07-02
Add deprovision logic for dedicated WAF CDN plan

markdboyd:

- Add deprovision pipeline and API handling for dedicated WAF CDN plan
- Add integration tests for deprovisioning dedicated WAF CDN plan instances
- Move common deprovision tests to `tests/integration/test_deprovision_common_checks.py` using pytest parameterization
- Moved shared deprovision test helpers to `tests/lib/cdn/deprovision.py`

## Things to check

## Security considerations

There should be no security considerations to these code changes directly. However, having a CDN with dedicated WAF will be a security benefit to customers.


broker/models.py
```diff
@@ -173,6 +173,8 @@ def __repr__(self):
 
 class CDNDedicatedWAFServiceInstance(CDNServiceInstance):
     dedicated_waf_web_acl_arn = mapped_column(db.String)
+    dedicated_waf_web_acl_id = mapped_column(db.String)
+    dedicated_waf_web_acl_name = mapped_column(db.String)
```
bengerman13, line 177: there should be a migration for this, right?

markdboyd: Yeah, I forgot that those exist

bengerman13 (approved):

bengerman13 (approved):

PR #340, opened 2024-07-03
fix tests for retries on deleting WAF web ACL & move file

markdboyd:

- fix tests for retries on deleting WAF web ACL to mock API calls the correct number of times
- update the delete web ACL code to retry 10 times before giving up 
- move file since it is an integration and not a unit test

## Things to check

## Security considerations

None


JasonTheMain (approved):
Looks good, limit is true 10 now

PR #341, opened 2024-07-05
remove secrets file from pipeline in favor of Credhub

markdboyd:

- remove secrets file from pipeline in favor of Credhub. Secrets for this pipeline configuration will now be read from Credhub

## Things to check

## Security considerations

Credhub is one of our approved secret stores. This PR is a step towards improving consistency for all of our pipelines to store their secrets in Credhub.


JasonTheMain (approved):
Nice removal of secrets

PR #343, opened 2024-07-08
Fix pipeline and address security issue

markdboyd:

- [upgrade to certifi>=2024.07.04](https://github.com/cloud-gov/external-domain-broker/commit/ae89f337cc92fc74151b593fbebf7f361e0c902f)
- Fix issue with Credhub variables in pipeline definition

## Things to check

## Security considerations

PR addresses open security issues identified by Dependabot


JasonTheMain (approved):
Looks good to me Mark

PR #344, opened 2024-08-01
Add update pipeline for with CDN dedicated WAF plan

markdboyd:

- Add pipeline of tasks for updating a CDN with dedicated WAF instance
- Add tasks to handle updating web ACL, route 53 health checks, and shield associated health checks
- Update integration tests
- Rename `route_53_health_check_ids` field to `route53_health_checks`

## Things to check

## Security considerations

None, just providing the correct logic to update a CDN with dedicated WAF instance


tests/integration/cdn_dedicated_waf/test_cdn_dedicated_waf_provisioning.py
```diff
@@ -203,3 +351,57 @@ def subtest_provision_associates_health_checks(tasks, shield, instance_model):
...
+
+def subtest_updates_associated_health_checks(tasks, shield, instance_model):
+    db.session.expunge_all()
+    service_instance = db.session.get(instance_model, "4321")
+    if not service_instance:
+        raise Exception("Could not load service instance")
+
+    # get protection ID from initial creation
+    protection_id = service_instance.shield_associated_health_checks[0]["protection_id"]
+
```
rcgottlieb, line 364: Does the commented-out code below need to be here?

rcgottlieb (approved):

PR #345, opened 2024-08-08
Add upgrade path from CDN plan to CDN with dedicated WAF plan

markdboyd:

- Add pipeline to upgrade `CDNServiceInstance` instances to `CDNDedicatedWAFServiceInstance` instances
- Add integration tests for upgrade path
- Refactor logic for handling domain updates in API `update` logic into a separate helper and add unit tests

## Things to check

## Security considerations

Should be no direct security impact of these changes, but they will allow CDN customers to move to CDNs with a custom WAF associated, which would be beneficial to their security and overall platform security


tests/unit/test_provision.py
markdboyd, line 1: Contents moved into `tests/unit/test_utils.py`

bengerman13 (approved):

broker/lib/utils.py
```diff
@@ -0,0 +1,78 @@
...
+        validators.UniqueDomains(domain_names).validate(instance)
+        # If domain names have not changed, then there is no need for a new certificate
+        no_domain_updates = no_domain_updates and (
+            sorted(domain_names) == sorted(instance.domain_names)
+        )
+        instance.domain_names = domain_names
+
+    if is_cdn_instance(instance) and no_domain_updates:
+        logger.info("domains unchanged, no need for new certificate")
+        instance.new_certificate = instance.current_certificate
```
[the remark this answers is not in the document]
markdboyd: I agree. I kind of dislike mutating the state of input parameters in general, because it leads to unpredictable behavior. How about 9a2b2b714d26f3ab0470680f15f115d75f1dc83a ?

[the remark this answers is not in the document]
bengerman13: yeah, that looks better

bengerman13 (dismissed):

broker/lib/utils.py
```diff
@@ -56,18 +56,16 @@ def parse_domain_options(params):
         return [d.strip().lower() for d in domains]
 
 
-def handle_domain_updates(params, instance):
-    domain_names = parse_domain_options(params)
-    no_domain_updates = True
-    if domain_names is not None:
+def handle_domain_updates(params, instance) -> list[str]:
+    updated_domain_names = parse_domain_options(params) or []
+    if len(updated_domain_names) > 0:
```
bengerman13, line 61: `parse_domain_options` returns `None` if there are no domain options passed in - I think using `None` as our no update case would be a little better here, if only so we have one fewer place where we change the value

broker/api.py
```diff
@@ -294,18 +294,17 @@ def update(  # noqa C901 # TODO: simplify this function
         if instance.has_active_operations():
             raise errors.ErrBadRequest("Instance has an active operation in progress")
 
-        (updated_domain_names, no_domain_updates) = handle_domain_updates(
-            params, instance
-        )
+        updated_domain_names = handle_domain_updates(params, instance)
```
bengerman13, line 297: tiny nit - to me `updated_domain_names` reads like a delta, or a list of the names that have changes. Maybe `new_domain_names` or `requested_domain_names`?

tests/unit/test_utils.py
```diff
@@ -146,9 +141,96 @@ def test_cdn_handle_domain_updates_with_changes(
...
+        factories.CDNDedicatedWAFServiceInstanceFactory,
+        factories.DedicatedALBServiceInstanceFactory,
+    ],
+)
+def test_handle_domain_errors_on_missing_dns(
+    no_context_clean_db, no_context_app, dns, factory
+):
+    with no_context_app.app_context():
+        dns.add_cname("_acme-challenge.moo.com")
+        domain_names = ["moo.com", "cow.com"]
```
bengerman13, line 165: We should use only official example domains or domains we control for test data - can we use something like `moo.example.com` and `cow.example.net`

bengerman13 (changes requested):

bengerman13 (approved):

PR #347, opened 2024-08-12
rename isntance variable for readability

bengerman13:

- rename isntance variable for readability

## Things to check

## Security considerations

None


markdboyd (approved):
I don't think this is required. It was just my confusion. But 👍

markdboyd:
Closing in favor of #348

PR #348, opened 2024-08-12
Update dedicated ALB logic and add more unit tests

markdboyd:

- Refactor logic for `get_lowest_dedicated_alb` into separate functions 
- Add unit tests
- Rename variables / add comments to clarify code behavior

## Things to check

## Security considerations

None, just refactoring code for clarity

### DIFF
--- a/broker/tasks/alb.py
+++ b/broker/tasks/alb.py
@@ -40,9 +40,11 @@ def get_lowest_used_alb(listener_arns) -> tuple[str, str]:
     return listener_data["LoadBalancerArn"], listener_data["ListenerArn"]
 
 
-def get_lowest_dedicated_alb(service_instance, db):
+def get_potential_listeners(service_instance):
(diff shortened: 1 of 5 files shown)


broker/tasks/alb.py
```diff
@@ -36,9 +40,11 @@ def get_lowest_used_alb(listener_arns) -> tuple[str, str]:
     return listener_data["LoadBalancerArn"], listener_data["ListenerArn"]
 
 
-def get_lowest_dedicated_alb(service_instance, db):
+def get_potential_listeners(service_instance):
```
bengerman13, line 43:
```suggestion
def get_potential_listeners_for_dedicated_instance(service_instance):
```

I think it reads a bit better if we spell out that this function is for dedicated instances
(that makes me wonder if this should be an instance method, too, but that's probably out of scope)

bengerman13 (approved):

PR #349, opened 2024-08-19
Update web ACL creation to use a predictable name

markdboyd:

Related to https://github.com/cloud-gov/cg-provision/pull/1735

- Update web ACL creation logic to use a predictable name so that IAM permissions for managing these resources can be tightly scoped

## Things to check

## Security considerations

These changes allow for more tightly scoped IAM permissions for creating/deleting WAF web ACLs. See https://github.com/cloud-gov/cg-provision/pull/1735.

### DIFF
--- a/broker/config.py
+++ b/broker/config.py
@@ -57,6 +57,8 @@ class Config:
         # the maximum from AWS is 25, and we alert when we have 20 on a given alb
         self.MAX_CERTS_PER_ALB = 19
 
+        self.DEDICATED_WAF_NAME_PREFIX = f"cg-external-domains-{self.FLASK_ENV}"
+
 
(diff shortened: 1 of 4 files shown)


bengerman13 (approved):
LGTM, although worth confirming the length constraints if you haven't already

markdboyd:
[Yeah, the length constraint for a web ACL name is 128 characters.](https://docs.aws.amazon.com/waf/latest/APIReference/API_WebACL.html)

And I confirmed in Python that our naming scheme should not exceed 128 characters in the worst case scenario:

```
>>> foo = f"cg-external-domains-production-{uuid.uuid4()}-dedicated-waf"
>>> len(foo)
81
```

PR #350, opened 2024-08-21
Add acceptance tests for CDN with dedicated WAF lan

markdboyd:

- Add acceptance tests for CDN with dedicated WAF plan
- Refactor pipeline to address YAML linter errors

## Things to check

## Security considerations

None, just adding acceptance tests for a new broker feature


JasonTheMain (approved):
Looks great Mark!!!

PR #351, opened 2024-08-21
Add tags

markdboyd:

- Add standard set of tags to all AWS resources created by the broker:
  - CloudFront distributions
  - IAM server certificates
  - WAF web ACLs
  - Route53 health checks
- Implement library to get name of CF org and space for use as tags
- Add unit tests for new tagging functionality
- Update integration tests to assert expected tag values

## Things to check

## Security considerations

Adding tags to deployed resources should have no impact on security


broker/lib/cf.py
```diff
@@ -0,0 +1,62 @@
...
+from urllib.parse import urljoin
+
+from broker.extensions import config
+
+
+class CFAPIClient:
+    _access_token: str
+    _access_token_expiration: float
+
+    def set_access_token(self):
```
bengerman13, line 13: `set_<x>` reads like I should be passing a value to this method - maybe this should be `fetch_access_token`?

broker/lib/tags.py
```diff
@@ -1,11 +1,85 @@
...
+from broker.lib.cf import CFAPIClient
+
+from enum import Enum
+from openbrokerapi.service_broker import (
+    ProvisionDetails,
+    ServicePlan,
+    Service,
+)
+
+cf_api_client = CFAPIClient()
```
bengerman13, line 13: my gut says this belongs in `extensions` to head off potential circular dependencies in the future

broker/lib/tags.py
```diff
@@ -1,11 +1,85 @@
...
+    if len(plans) > 1:
+        raise RuntimeError(
+            f"Found multiple plans for the given plan ID {details.plan_id}"
+        )
+    return create_resource_tags(
+        generate_tags(instance_id, catalog.name, plans[0], details)
+    )
+
+
+def create_resource_tags(tags: typing.Dict[str, str]) -> list[Tag]:
```
bengerman13, line 55: n.b, `typing.Dict` is deprecated - we can just use `dict`

broker/lib/tags.py
```diff
@@ -1,11 +1,85 @@
...
+    resource_tags = []
+    for tag_key in tags:
+        tag_value = tags[tag_key]
+        resource_tags = add_tag(resource_tags, tag_key, tag_value)
```
bengerman13, line 59:
```suggestion
    for tag_key, tag_value in tags.items():
        resource_tags = add_tag(resource_tags, tag_key, tag_value)
```

broker/lib/tags.py
```diff
@@ -1,11 +1,85 @@
...
+    Value: str
+
+
+def add_tag(tags: list[Tag], tag_key: str, tag_value: str) -> list[Tag]:
     if not tags:
-        tags = {}
-    tags["Items"] = tags.get("Items", [])
-    tags["Items"].append(
+        tags = []
+    tags.append(
```
bengerman13, line 29: is it OK if we get multiple `tag`s with the same `Key`?

tests/integration/test_cdn_provision_common_checks.py
```diff
@@ -8,15 +8,29 @@
...
 
 
+# @pytest.fixture
+# def provisioned_instance_client(
+#     client, dns, organization_guid, space_guid, instance_model
+# ):
+
+#     yield client
+
+
```
bengerman13, line 18: do we still need this?

broker/lib/tags.py
markdboyd: I don't think I can put this in `extensions`, because `cf.py` which contains `CFAPIClient` itself loads `config` from `extensions`. So trying to load `CFAPIClient` from `extensions` would create a circular dependency on itself

broker/lib/tags.py
markdboyd: Good point. f094f49e1f147646b17794a528e5c7d6986bc33a should address it

bengerman13 (approved):

PR #352, opened 2024-08-23
Fix tests

markdboyd:

- Make sure boto client for WAF is scoped to correct region
- Fix token refresh logic and update unit tests
- Add new helper for integration tests

## Things to check

## Security considerations

These changes are bug fixes with no expected security impact


bengerman13 (approved):

broker/lib/cf.py
```diff
@@ -42,7 +47,8 @@ def access_token(self):
     def is_token_expiring(self):
         if self._access_token_expiration:
             now_utc = datetime.datetime.now(datetime.timezone.utc)
-            if now_utc.timestamp() - self._access_token_expiration <= 30:
+            if self._access_token_expiration - now_utc.timestamp() <= 30:
```
markdboyd, line 50: With the previous logic, it was always recognizing the token as expired. Let's say the token expires 100 seconds from now. In that case, this math:

```
now_utc.timestamp() - self._access_token_expiration
```

would give you roughly `-100`. And since `-100 <= 30`, it will treat the token as expired and refresh it.

However, a token that expires 100 seconds from now should not be refreshed. So changing the logic to 

```
self._access_token_expiration - now_utc.timestamp()
```

means that the result should be `100`. And since `100 > 30`, the token will not be refreshed

PR #355, opened 2024-08-27
Fix more provision errors

markdboyd:

- Fix logic for setting caller reference name when creating Route53 health checks to respect length limit
- Add logic for handling case where IAM server certificate was already uploaded but not persisted to the database. Add integration test for this case

## Things to check

## Security considerations

None, just fixing bugs preventing resources from provisioning correctly


JasonTheMain (approved):
Looks Good

PR #357, opened 2024-08-29
Update shield health check logic

markdboyd:

- Update logic for associating Route53 health checks with Shield protection to account for the fact that **only one health check can be associated to a protection**
  - Update tests to account for the change
- Add integration tests for route53 and shield tasks

## Things to check

## Security considerations

None, just updating broker behavior


tests/integration/cdn_dedicated_waf/update.py
```diff
@@ -19,6 +20,41 @@ def subtest_updates_health_checks(
     route53.expect_create_health_check(service_instance.id, "bar.com", 0)
     route53.expect_change_tags_for_resource("bar.com", service_instance.tags)
 
+    # delete_health_check = [
+    #     check
+    #     for check in service_instance.route53_health_checks
+    #     if check["domain_name"] == "example.com"
+    # ][0]
+    # route53.expect_delete_health_check(delete_health_check["health_check_id"])
+
```
jameshochadel, line 29: Probably want to fully remove these lines?

jameshochadel (dismissed):
Looks fine to me, but I'm a surface-level reviewer on this one as I'm not super familiar with the codebase.

PR #359, opened 2024-08-29
Harden image

dandersonsw:

- Updates dockerfile to use our hardened base image
- Updates pipeline to use the hardened testing image

## Things to check

## Security considerations

Updating code to use hardened image

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -239,19 +239,9 @@ jobs:
     params: {depth: 1}
     trigger: true
     passed: [reconfigure]
-
-  - put: dev-docker-image
-    # We push this docker image _only_ so that we can use it in the next task
-    # below.  :shrug: concourse.
-    params:
-      build: src
-      dockerfile: src/docker/Dockerfile.dev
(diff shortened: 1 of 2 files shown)


ci/pipeline.yml
markdboyd, line 1: Can we split the PR to do this file change and the change to the Dockerfile separately? Once we merge this change, the pipeline is broken until the hardened image is available. Whereas the change to the Dockerfile can be done anytime

dandersonsw: I think the pipeline is going to break either way, because once the change to the Dockerfile is made, the pipeline is going to try to build the docker image, but it won't be able to because there isn't anything set to populate the base_image arg in the dockerfile.

markdboyd: OK. I might hold off on approving/merging until I get at least one more fix in for the pipeline then, if that's okay

dandersonsw: Yep, no problem. It might be good to wait and roll out a change like this next week anyways

markdboyd (approved):

PR #360, opened 2024-08-29
fix imports for huey_consumer

markdboyd:

- fix imports for huey_consumer based on https://github.com/cloud-gov/external-domain-broker/pull/358

## Things to check

## Security considerations

None, just fixing broker imports


bengerman13 (approved):
This looks like a valid fix. It might also make sense to either make an `__init__.py` to make the pipelines import cleaner, or use a glob here

PR #362, opened 2024-08-30
Fix logic waiting for distribution to be disabled

markdboyd:

We are getting this error when our acceptance tests run in CI:

```shell
botocore.errorfactory.DistributionNotDisabled: An error occurred (DistributionNotDisabled) when calling the DeleteDistribution operation: The distribution you are trying to delete has not been disabled., 
```

This error doesn't cause the tests or service deletion to ultimately fail, because it just retries until the distribution has been fully disabled and deletes it. However, it would be preferable to have the task which is supposed to wait for a distribution to be disabled, `wait_for_distribution_disabled`, to correctly wait long enough so that this error does not occur.

[The problem is that we need to wait not only for the status of the distribution to be disabled, but also for that change to have fully deployed before attempting the deletion](https://stackoverflow.com/questions/43077173/deleting-a-cloudfront-distribution-with-boto3).

This PR adds logic to wait_for_distribution_disabled to ensure that task waits until distribution is disabled and fully deployed before continuing. The PR also adds a small integration test to ensure that the new behavior is working as expected.

## Things to check

## Security considerations

None, just fixing a small bug in the broker behavior

### DIFF
--- a/broker/tasks/cloudfront.py
+++ b/broker/tasks/cloudfront.py
@@ -223,9 +223,9 @@ def wait_for_distribution_disabled(operation_id: int, **kwargs):
     if service_instance.cloudfront_distribution_id is None:
         return
 
-    enabled = True
+    distribution_disabled = False
     num_times = 0
-    while enabled:
+    while not distribution_disabled:
         num_times += 1
         if num_times >= 60:
             logger.info(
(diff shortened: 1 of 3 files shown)


JasonTheMain (approved):
Like the new test, very nice!

PR #369, opened 2024-09-05
Fix: only tag certs when we have tags to tag them with

bengerman13:

- only tag certs when we have tags to tag them with

## Things to check

## Security considerations

None

### DIFF
--- a/tests/lib/fake_iam.py
+++ b/tests/lib/fake_iam.py
@@ -29,7 +29,9 @@ class FakeIAM(FakeAWS):
             "ServerCertificateName": name,
             "Tags": tags,
         }
-        self.stubber.add_response(method, {}, request)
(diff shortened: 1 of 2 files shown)


tests/lib/fake_iam.py
```diff
@@ -29,7 +29,9 @@ def expect_tag_server_certificate(self, name: str, tags: list[Tag]):
             "ServerCertificateName": name,
             "Tags": tags,
         }
-        self.stubber.add_response(method, {}, request)
+        # sending an empty tags causes an error in IAM.
+        if tags != []:
```
bengerman13, line 33: I'm not 100% sure this is a clean solution here, but it works

markdboyd (approved):

PR #370, opened 2024-09-05
Fix: looser error code checking and log codes when it fails

bengerman13:

- use a substring check on error code, since we seem to be not matching right now
- log when our error code match fails, so we can see why we fail if this doesn't work

## Things to check

## Security considerations

None

### DIFF
--- a/broker/tasks/iam.py
+++ b/broker/tasks/iam.py
@@ -58,7 +58,7 @@ def upload_server_certificate(operation_id: int, **kwargs):
             response = get_response["ServerCertificate"]
         else:
             logger.error(
-                f"Got this coe uploading server certificate: {e.response['Error']}"
+                f"Got this code uploading server certificate: {e.response['Error']}"
             )
             raise e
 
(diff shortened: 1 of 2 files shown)


broker/tasks/iam.py
```diff
@@ -51,12 +51,15 @@ def upload_server_certificate(operation_id: int, **kwargs):
...
     except ClientError as e:
-        if e.response["Error"]["Code"] == "EntityAlreadyExistsException":
+        if "EntityAlreadyExists" in e.response["Error"]["Code"]:
             get_response = iam.get_server_certificate(
                 ServerCertificateName=certificate.iam_server_certificate_name,
             )
             response = get_response["ServerCertificate"]
         else:
+            logger.error(
+                f"Got this coe uploading server certificate: {e.response['Error']}"
```
markdboyd, line 61:
```suggestion
                f"Got this code uploading server certificate: {e.response['Error']}"
```

markdboyd (approved):

PR #371, opened 2024-09-05
Miscellaneous fixes

markdboyd:

- Make tags optional for creating WAF web ACL
- Make tags optional for creating health check alarms
- Handle deletion failures where certificates may already have been deleted by checking to see if the certificate exists before deleting

## Things to check

## Security considerations

None, just addressing observed bugs in the logs

### DIFF
--- a/broker/tasks/alb.py
+++ b/broker/tasks/alb.py
@@ -162,10 +162,6 @@ def add_certificate_to_alb(operation_id, **kwargs):
     service_instance = operation.service_instance
     certificate = service_instance.new_certificate
 
-    if not certificate:
-        logger.info(f"Could not find certificate for instance {service_instance.id}")
-        return
-
     operation.step_description = "Adding SSL certificate to load balancer"
     flag_modified(operation, "step_description")
     db.session.add(operation)
--- a/broker/tasks/alb.py
+++ b/broker/tasks/alb.py
@@ -162,6 +162,10 @@ def add_certificate_to_alb(operation_id, **kwargs):
     service_instance = operation.service_instance
     certificate = service_instance.new_certificate
 
+    if not certificate:
+        logger.info(f"Could not find certificate for instance {service_instance.id}")
+        return
+
     operation.step_description = "Adding SSL certificate to load balancer"
     flag_modified(operation, "step_description")
     db.session.add(operation)
--- a/broker/tasks/cloudwatch.py
+++ b/broker/tasks/cloudwatch.py
@@ -189,7 +189,7 @@ def _delete_alarms(service_instance, alarm_names_to_delete):
             )
         else:
             logger.error(
-                f"Got this code uploading server certificate: {e.response['Error']}"
+                f"Got this code deleting Cloudwatch alarms: {e.response['Error']}"
             )
             raise e
     service_instance.cloudwatch_health_check_alarms = [
--- a/broker/tasks/cloudwatch.py
+++ b/broker/tasks/cloudwatch.py
@@ -1,5 +1,6 @@
 import logging
 
+from botocore.exceptions import ClientError
 from sqlalchemy.orm.attributes import flag_modified
 
 from broker.aws import cloudwatch_commercial
@@ -176,11 +177,17 @@ def _create_health_check_alarm(health_check_id, tags) -> str:
 def _delete_alarms(service_instance, alarm_names_to_delete):
     try:
         cloudwatch_commercial.delete_alarms(AlarmNames=alarm_names_to_delete)
-    except cloudwatch_commercial.exceptions.ResourceNotFound:
-        logger.info(
-            "Cloudwatch alarms not found",
-            extra={"alarm_names": alarm_names_to_delete},
-        )
+    except ClientError as e:
+        if "ResourceNotFound" in e.response["Error"]["Code"]:
+            logger.info(
+                "Cloudwatch alarms not found",
+                extra={"alarm_names": alarm_names_to_delete},
+            )
+        else:
+            logger.error(
+                f"Got this code uploading server certificate: {e.response['Error']}"
+            )
+            raise e
     service_instance.cloudwatch_health_check_alarms = [
         health_check_alarm
         for health_check_alarm in service_instance.cloudwatch_health_check_alarms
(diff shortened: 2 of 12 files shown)


broker/tasks/alb.py
```diff
@@ -162,6 +162,10 @@ def add_certificate_to_alb(operation_id, **kwargs):
     service_instance = operation.service_instance
     certificate = service_instance.new_certificate
 
+    if not certificate:
```
markdboyd, line 165: I'm not sure if this is a good fix at all, but it is attempting to address this error which was observed in the logs

```
File "/home/vcap/app/broker/tasks/alb.py", line 172, in add_certificate_to_alb,     Certificates=[{"CertificateArn": certificate.iam_server_certificate_arn}],,                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^, AttributeError: 'NoneType' object has no attribute 'iam_server_certificate_arn'
```

broker/tasks/cloudwatch.py
```diff
@@ -176,11 +181,17 @@ def _create_health_check_alarm(health_check_id, tags) -> str:
...
-        )
+    except ClientError as e:
+        if "ResourceNotFound" in e.response["Error"]["Code"]:
+            logger.info(
+                "Cloudwatch alarms not found",
+                extra={"alarm_names": alarm_names_to_delete},
+            )
+        else:
+            logger.error(
+                f"Got this code uploading server certificate: {e.response['Error']}"
```
bengerman13, line 192:
```suggestion
                f"Got this code deleting Cloudwatch alarms: {e.response['Error']}"
```

broker/tasks/iam.py
```diff
@@ -135,18 +135,31 @@ def delete_previous_server_certificate(operation_id: str, **kwargs):
...
+    certificates_to_delete = Certificate.query.filter(
         and_(
             Certificate.service_instance_id == service_instance.id,
             Certificate.id != service_instance.current_certificate_id,
         )
-    ).all():
+    ).all()
+    for certificate in certificates_to_delete:
         try:
+            print(certificate.iam_server_certificate_name)
```
bengerman13, line 146: this should probably go through logging, right? or was it supposed to be deleted?

markdboyd: yeah, this was for debugging. i'll remove

broker/tasks/iam.py
```diff
@@ -135,18 +135,31 @@ def delete_previous_server_certificate(operation_id: str, **kwargs):
...
+            iam.get_server_certificate(
+                ServerCertificateName=certificate.iam_server_certificate_name
+            )
```
bengerman13, line 149: what's the goal here?

markdboyd: It's complicated. And more of a matter of IAM permissions. 

According to AWS support, the only way to make a DeleteServerCertificate operation return NoSuchEntity error instead of AccessDenied is to give iam:DeleteServerCertificate on all certificates (*), rather than certificates under the path managed by this broker. However, these permissions would allow this domain broker to delete certificates managed by other brokers, so we do not want to do that.

Instead, we have updated the delete certificate code to check if the certificate exists before attempting the deletion. If the certificate no longer exists, then we do not attempt the deletion.

markdboyd: I guess the real underlying goal was to address errors like this in th elogs

```
external-domain-broker-gov-production is not authorized to perform: iam:DeleteServerCertificate on resource: server certificate <redacted> because no identity-based policy allows the iam:DeleteServerCertificate action, 
```

broker/tasks/alb.py
bengerman13: I think this will silence that error, but cause another one down the line when we try to do something with e.g, the cert arn

markdboyd: OK. yeah, I'm not sure what else to do, because it's hard to trace the original cause

markdboyd: I'm not seeing these errors right now, so I'll remove this possibly unhelpful defensive programming

broker/tasks/iam.py
markdboyd: I inverted the control flow so that we only do a `get_server_certificate` call if the `delete_server_certificate` fails with `AccessDenied`, which might happen when the certificate is already deleted. The refactor preserves the existing behavior much better and just catches the undesirable error case

92728c872c559e8658e62743efb9b1f0bc94c556

broker/tasks/iam.py
```diff
@@ -135,18 +135,38 @@ def delete_previous_server_certificate(operation_id: str, **kwargs):
...
-    ).all():
+    ).all()
+    for certificate in certificates_to_delete:
         try:
             iam.delete_server_certificate(
                 ServerCertificateName=certificate.iam_server_certificate_name
             )
-        except iam_commercial.exceptions.NoSuchEntityException:
-            pass
+        except ClientError as e:
```
markdboyd, line 149: The nested try/catch here is pretty ugly, but I'm not sure how else to handle it

broker/tasks/iam.py
bengerman13: ah, that makes more sense to me now

broker/tasks/iam.py
bengerman13: I think it might be tidier:

```
try:
    get_response = iam.get_server_certificate(...)
except ClientError as e:
    if "NoSuchEntity" in e.response["Error"]["Code"]:
        return # the cert is gone, we've got nothing to do
# now we know we can see the cert, so any errors should be treated as unexpected
iam.delete_server_certificate(...)
```

n.b. I put a `return` here - that assumes pulling this logic out into a function (which I think is probably a good idea). You could use `next` instead if you wanted to leave the loop

broker/tasks/cloudwatch.py
```diff
@@ -137,6 +136,10 @@ def _create_health_check_alarm(health_check_id, tags) -> str:
     alarm_name = _get_alarm_name(health_check_id)
 
     # create alarm
+    kwargs = {}
+    if tags is not None:
```
bengerman13, line 140: is it possible we'll get into a state where `service_instance.tags` == `[]`?
if so, we should change this to `if tags:`

bengerman13 (approved):

bengerman13 (approved):

PR #372, opened 2024-09-06
Refactor code

markdboyd:

- Refactor logic for deleting server certificates to check if they exist first, then only attempt deletion if the certificate does exist
- Moved common test fixtures for identifiers to shared lib
- Update `expect_get_server_certificate` on fake IAM service to use dummy values for cert and chain, which makes testing much less cumbersome

## Things to check

## Security considerations

None, just refactoring code for clarity and maintainability


JasonTheMain (approved):
Looks Good.

PR #375, opened 2024-09-09
More fixes

markdboyd:

- Update pipeline to use (hopefully correct) certificate for SMTP 
- Remove unnecessary fixture data for tests
- Update `remove_certificate_from_previous_alb` to verify removal of certificate from listener and add integration tests. This change addresses observed errors in the logs where attempts to delete the certificate encounter a `DeleteConflict` error because the deletion attempt before the certificate has finished being removed from the listener
- Update logic for getting expired certificates to join the service instance table on `ServiceInstance.current_certificate_id == Certificate.id` instead of `ServiceInstance.id == Certificate.service_instance_id`. Joining on `current_certificate_id` ensures that we are only checking whether the **current certificate is expiring**, rather than checking whether any prevoius certificate connected to the instance is expiring/expired.
  - Add integration tests to verify the new behavior

## Things to check

## Security considerations

None, just addressing various observed bugs


broker/tasks/cron.py
```diff
@@ -31,15 +32,36 @@
...
+            < datetime.datetime.now()
+        )
+    )
+    results = db.session.execute(query).fetchall()
+    certificates = []
+    for result in results:
+        [certificate_id] = result
+        certificate = db.session.get(Certificate, certificate_id)
+        certificates.append(certificate)
+    return certificates
```
bengerman13, line 57: I think this will fix the problem we're seeing, but we're now running a lot of queries here:
- one query to get the list of certs
- one query _per cert_ to get the cert (line 55)


I think it would be better to have the query return the service instances instead and pass those straight to `instances` on line 65

we're also filtering twice for the service being active (line 45 and line 68) - I think it's an improvement to move that check to 45, but we should drop it on 68 as well

(not your fault, but while you're deleting duplicate checks: we're looking for active operations twice: line 69 and 75. We can get rid of one of those)

bengerman13 (approved):

bengerman13 (approved):

PR #376, opened 2024-09-10
Fix route53 create health check behavior on instance type update

markdboyd:

- Fix error observed in logs during `create_health_checks` task when upgrading CDN to CDN with dedicated WAF instances
- Fix other integration test failures

## Things to check

## Security considerations

None, just fixing bugs

### DIFF
--- a/broker/tasks/route53.py
+++ b/broker/tasks/route53.py
@@ -247,7 +247,7 @@ def create_health_checks(operation_id: int, **kwargs):
     created_health_checks = _create_health_checks(
         service_instance,
         service_instance.domain_names,
-        service_instance.route53_health_checks,
+        [],
     )
     service_instance.route53_health_checks = created_health_checks
     flag_modified(service_instance, "route53_health_checks")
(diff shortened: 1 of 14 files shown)


broker/tasks/route53.py
```diff
@@ -247,7 +247,7 @@ def create_health_checks(operation_id: int, **kwargs):
     created_health_checks = _create_health_checks(
         service_instance,
         service_instance.domain_names,
-        service_instance.route53_health_checks,
+        [],
```
markdboyd, line 250: This is an attempt to fix this error which was observed in the logs:

```
 File "broker/tasks/route53.py", line 247, in create_health_checks,     
  created_health_checks = _create_health_checks(,                             
  ^^^^^^^^^^^^^^^^^^^^^^,   
  File "broker/tasks/route53.py", line 377, in _create_health_checks,     
    updated_health_checks.append(,     
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^, AttributeError: 'NoneType' object has no attribute 'append', 
```

However, I'm not confident it will actually resolve it because I can't reproduce the error in a local integration test

bengerman13 (approved):

bengerman13 (approved):

PR #379, opened 2024-09-11
Fix idempotency

markdboyd:

- [delete route53.create_health_checks method and replace with route53.create_new_health_checks & refactor tests](https://github.com/cloud-gov/external-domain-broker/commit/48a26aced57ff69e6412818fc32cf9aa8b970267) 
- [update update_health_check_alarms to handle missing cloudwatch_health_check_alarms property on service instance](https://github.com/cloud-gov/external-domain-broker/commit/1d8694d786fb4d31f99563792160406a35c00771) 
- [add test for update_health_check_alarms idempotency](https://github.com/cloud-gov/external-domain-broker/commit/63c9f38bb0f1c70f30af06bbb444e9a8e8249f0c)

## Things to check

## Security considerations

None, just fixing bugs with broker idempotency


JasonTheMain (approved):
looks great mark!!

PR #380, opened 2024-09-11
Add cron pipelines dedicated WAF

markdboyd:

- Add missing code for restarting stalled pipelines for CDN with dedicated WAF instances
- Update code to use enum for service instance types 
- Catch `IndexError` when restarting operations
- Add integration test for CDN with dedicated WAF renewals

## Things to check

## Security considerations

None, just fixing bugs


broker/models.py
```diff
@@ -19,6 +19,14 @@
 logger = logging.getLogger(__name__)
 
 
+class ServiceInstanceTypes(Enum):
```
bengerman13, line 22: :chef-kissing-fingers:

bengerman13 (approved):

PR #383, opened 2024-09-12
Update delete tasks to handle empty values

markdboyd:

- Update various delete tasks to ensure they handle empty values for the relevant columns correctly
- Add/update tests as necessary to ensure that tasks handle empty values correctly

## Things to check

## Security considerations

None, just fixing broker bugs

### DIFF
--- a/broker/tasks/shield.py
+++ b/broker/tasks/shield.py
@@ -110,7 +110,7 @@ def disassociate_health_check(operation_id: int, **kwargs):
     service_instance = operation.service_instance
 
     shield_associated_health_check = service_instance.shield_associated_health_check
-    if shield_associated_health_check == None or shield_associated_health_check == {}:
+    if shield_associated_health_check is None or shield_associated_health_check == {}:
         logger.info("No health check to disassociate from Shield")
         return
 
--- a/broker/tasks/shield.py
+++ b/broker/tasks/shield.py
@@ -110,7 +110,7 @@ def disassociate_health_check(operation_id: int, **kwargs):
     service_instance = operation.service_instance
 
     shield_associated_health_check = service_instance.shield_associated_health_check
-    if shield_associated_health_check == None:
+    if shield_associated_health_check == None or shield_associated_health_check == {}:
         logger.info("No health check to disassociate from Shield")
         return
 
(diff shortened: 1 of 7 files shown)


broker/tasks/shield.py
```diff
@@ -110,7 +110,7 @@ def disassociate_health_check(operation_id: int, **kwargs):
     service_instance = operation.service_instance
 
     shield_associated_health_check = service_instance.shield_associated_health_check
-    if shield_associated_health_check == None:
+    if shield_associated_health_check == None or shield_associated_health_check == {}:
```
bengerman13, line 113:
```suggestion
    if shield_associated_health_check is None or shield_associated_health_check == {}:
```

broker/tasks/waf.py
```diff
@@ -89,6 +89,7 @@ def delete_web_acl(operation_id: str, **kwargs):
     if (
         not service_instance.dedicated_waf_web_acl_name
         or not service_instance.dedicated_waf_web_acl_id
+        or not service_instance.dedicated_waf_web_acl_arn
```
bengerman13, line 92: so there are instances where we will have a name and id for the acl, but no arn, and that means ignoring the delete is fine? Or an id and an acl but no name?

tests/integration/test_cloudwatch.py
```diff
@@ -641,6 +652,12 @@ def test_delete_health_check_alarms_unmigrated_instance(
     unmigrated_cdn_dedicated_waf_service_instance_operation_id,
     cloudwatch_commercial,
 ):
+    service_instance = clean_db.session.get(
+        CDNDedicatedWAFServiceInstance,
+        service_instance_id,
+    )
+    assert service_instance.cloudwatch_health_check_alarms == None
```
bengerman13, line 659:
```suggestion
    assert service_instance.cloudwatch_health_check_alarms is None
```

bengerman13 (changes requested):

broker/tasks/waf.py
markdboyd: All of these properties are set at the same time by the task that creates a web ACL

https://github.com/cloud-gov/external-domain-broker/blob/main/broker/tasks/waf.py#L72-L74

So in theory, if one of them missing, they all would be, but it still felt safer to test for all three. I noticed I wasn't checking all three properties, so I added this last one

But I'm open to any way of handling this better

bengerman13: I think it makes the most sense to check only for the value(s) that we'll need to do the delete (although since they're all set in the same place, it probably doesn't matter _that_ much)

bengerman13 (approved):

PR #384, opened 2024-09-13
Update delete certificate logic

markdboyd:

- Add logic to check for certificate before attempting deletion. Skip deletion if certificate does not exist
- Refactor tests to use shared helpers 
- Add more tests for IAM delete certificate idempotency
- Update IAM idempotency tests to use parameters for all service instance types

## Things to check

## Security considerations

None, just improving IAM certificate removal logic


JasonTheMain (approved):
Nice way to consolidate the delete action

Issue #291, opened 2024-03-18
During update to domain plan - old DNS records are not removed


ChrisMcGowan:
Plan: domain
Action: update-service to reduce the number of domains
Result: new cert is created and elb and DNS records for remaining domains are updated but removed domain DNS records for `_acme-challenge` and `<domain>` on the `external-domains->env>` domain are not removed 

## Test notes

- pushed a test ruby hello world
- Using the [cloud-gov-test-domain-3.com](http://cloud-gov-test-domain-3.com/) domain in commercial route 53 I created the needed _acme-challenge. and name records using the published CNAME instructions
- cf create-service external-domain domain my-domain -c '{"domains": "[pickles.cloud-gov-test-domain-3.com](http://pickles.cloud-gov-test-domain-3.com/),[mustard.cloud-gov-test-domain-3.com](http://mustard.cloud-gov-test-domain-3.com/)"}'
- After a few minutes I seen the _acme-challenge records on the [external-domains-production.cloud.gov](http://external-domains-production.cloud.gov/) hosted zone and then the CNAME records for the domain instance on production-domains3
- Added the domains to the org and mapped my routes
- Tested and pickles was the cert subject and pickles and mustard where both listed on SAN

- `cf update-service my-domain -c '{"domains": "[pickles.cloud-gov-test-domain-3.com](http://pickles.cloud-gov-test-domain-3.com/)"}'`
- I see the pickles record moved from production-domains-3 to production-domains-2.
- The old cert is now gone off production-domains-3
- pickles works and only has the subject

- Noticed the `_acme-challenge` and `<domain>.external-domains-<env>.cloud.gov` records for the removed domain still exist.  If someone goes to `<domain>.external-domains-<env>.cloud.gov` they will get the default `*.fr.cloud.gov` cert from that elb instead of a DNS failure as expected if the records are removed

## Acceptance Criteria

- [ ] Update logic on update to plans where a domain(s) are removed.

bengerman13:
I'm like 90% sure that if this is happening for the `domain` plan, it's also happening for `domain-with-cdn`

markdboyd:
Confirmed this is fixed by following the manual testing steps in the ticket description

PR #385, opened 2024-09-17
Remove old DNS records on update

markdboyd:

Fixes https://github.com/cloud-gov/external-domain-broker/issues/291

- Add `route53.remove_old_DNS_records` to remove TXT/ALIAS records in Route53 for domain names that are no longer active during an update operation
- Add `route53.remove_old_DNS_records` step to update pipelines for all service instances
- Add integration tests for `route53.remove_old_DNS_records` task
- Add step to integration tests of service update pipelines for `route53.remove_old_DNS_records`
- Clean up duplicate fixture declarations

## Things to check

## Security considerations

Should be none. Just making sure we clean up unnecessary DNS records in Route53 on update

### DIFF
--- a/broker/pipelines/cdn_dedicated_waf.py
+++ b/broker/pipelines/cdn_dedicated_waf.py
@@ -79,6 +79,7 @@ def queue_all_cdn_dedicated_waf_update_tasks_for_operation(
         .then(letsencrypt.initiate_challenges, operation_id, **correlation)
         .then(route53.create_TXT_records, operation_id, **correlation)
         .then(route53.wait_for_changes, operation_id, **correlation)
+        .then(route53.remove_old_TXT_records, operation_id, **correlation)
         .then(letsencrypt.answer_challenges, operation_id, **correlation)
         .then(letsencrypt.retrieve_certificate, operation_id, **correlation)
         .then(iam.upload_server_certificate, operation_id, **correlation)
--- a/broker/pipelines/cdn_dedicated_waf.py
+++ b/broker/pipelines/cdn_dedicated_waf.py
@@ -79,7 +79,7 @@ def queue_all_cdn_dedicated_waf_update_tasks_for_operation(
         .then(letsencrypt.initiate_challenges, operation_id, **correlation)
         .then(route53.create_TXT_records, operation_id, **correlation)
         .then(route53.wait_for_changes, operation_id, **correlation)
-        .then(route53.remove_old_TXT_records, operation_id, **correlation)
+        .then(route53.remove_old_DNS_records, operation_id, **correlation)
         .then(letsencrypt.answer_challenges, operation_id, **correlation)
         .then(letsencrypt.retrieve_certificate, operation_id, **correlation)
         .then(iam.upload_server_certificate, operation_id, **correlation)
--- a/broker/tasks/route53.py
+++ b/broker/tasks/route53.py
@@ -92,6 +92,29 @@ def remove_TXT_records(operation_id: int, **kwargs):
 
 
 @huey.retriable_task
+def remove_old_TXT_records(operation_id: int, **kwargs):
+    operation = db.session.get(Operation, operation_id)
+    service_instance = operation.service_instance
+
+    operation.step_description = "Removing old DNS TXT records"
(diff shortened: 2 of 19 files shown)


broker/tasks/route53.py
```diff
@@ -63,32 +63,31 @@ def remove_TXT_records(operation_id: int, **kwargs):
...
+    challenges = current_cert.challenges.all()
+
+    challenges_to_remove = [
+        challenge
+        for challenge in challenges
+        if challenge.domain not in service_instance.domain_names
+    ]
+    for challenge in challenges_to_remove:
+        _delete_TXT_record(challenge)
+        _delete_ALIAS_record(challenge.domain, service_instance)
```
markdboyd, line 90: I decided to delete both the old TXT and ALIAS records in the same task because it's only possible to detect what ALIAS record(s) to delete using the same logic that is used to delete old TXT records

broker/pipelines/cdn_dedicated_waf.py
```diff
@@ -79,6 +79,7 @@ def queue_all_cdn_dedicated_waf_update_tasks_for_operation(
         .then(letsencrypt.initiate_challenges, operation_id, **correlation)
         .then(route53.create_TXT_records, operation_id, **correlation)
         .then(route53.wait_for_changes, operation_id, **correlation)
+        .then(route53.remove_old_DNS_records, operation_id, **correlation)
```
markdboyd, line 82: I did have some concern about where to put this task in the pipeline. Technically, after this step is complete, ALIAS records for the old domains from the service will no longer work. And the new ALIAS records for the updated domains aren't created until later in the pipeline.

So if all of the domains for a CDN service were changed, there would be downtime between when the old DNS records were removed and the new ones were created. I decided that this is okay, since if you're updating all the domains for an instance, you might expect some downtime.

Thoughts?

broker/tasks/route53.py
```diff
@@ -414,3 +374,77 @@ def _delete_health_check(health_check_id):
...
+                                "HostedZoneId": service_instance.route53_alias_hosted_zone,
+                                "EvaluateTargetHealth": False,
+                            },
+                        },
+                    },
+                ]
+            },
+            HostedZoneId=config.ROUTE53_ZONE_ID,
+        )
+    except:  # noqa E722
```
markdboyd, line 416: I left this error handling in, but it seems quite aggressive. It seems like we do want to know if we get an error because of missing IAM permissions or something similar

broker/pipelines/cdn_dedicated_waf.py
bengerman13: changing _all_ the domains on an instance seems weird to me. 
I *think* the order doesn't matter too much, but maybe it makes the most sense either just before or just after removing the old cert (since that's where a request to the old domain is going to start breaking no matter what)

markdboyd: In this pipeline, `update_distribution` is what changes the value of `current_certificate`. So you think this task should happen after that?

broker/tasks/route53.py
bengerman13: I think the reasoning was that abandoned records are pretty harmless (same as why we don't track the change IDs and just assume Route53 will delete them successfully)

markdboyd: The available exceptions for this request don't seem to make it very easy to detect different classes of errors

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53/client/change_resource_record_sets.html

markdboyd: I guess overall, I'd like to be cleaning up Route53 vs leaving irrelevant records in there. As is, we have no way of knowing if that is working or if it is failing silently. So all of the changes in this PR could do nothing and we'd have no way of knowing

bengerman13 (approved):

broker/pipelines/cdn_dedicated_waf.py
bengerman13: seems like right around there would be the most perfect time, but I also don't think it matters a whole lot. I think the expectation should be once you call update-service removing that domain, the domain will no longer reliably work

PR #386, opened 2024-09-19
Add ability to subscribe to DDoS detected alarms

markdboyd:
Related to https://github.com/cloud-gov/private/issues/1097

- Add new `alarm_notification_email` that customers must specify for CDN w/dedicated WAF instances to receive notifications from Cloudwatch alarms
- Add new Cloudwatch alarm when DDoS is detected on the protected CDN
- Add dedicated SNS topic for sending notifications to the customer
- Ensure that SNS topic and DDoS alarm are created when migrating CDN to CDN w/dedicated WAF plan
- Add/update tests

## Things to check

## Security considerations

No direct security considerations for the broker, but these changes will facilitate better security monitoring for customers


broker/lib/cdn.py
```diff
@@ -1,8 +1,14 @@
 from broker.models import ServiceInstanceTypes
 
 
-def is_cdn_instance(service_instance):
+def is_cdn_instance(service_instance) -> bool:
     return service_instance.instance_type in [
         ServiceInstanceTypes.CDN.value,
         ServiceInstanceTypes.CDN_DEDICATED_WAF.value,
     ]
```
bengerman13, line 8: maybe late to be checking this, but why aren't we just using `isinstance(service_intance, CDNServiceInstance)`

broker/lib/cdn.py
```diff
@@ -1,8 +1,14 @@
...
+
+
+def is_cdn_dedicated_waf_instance(service_instance) -> bool:
+    return (
+        service_instance.instance_type == ServiceInstanceTypes.CDN_DEDICATED_WAF.value
+    )
```
bengerman13, line 14: same question here: why not just `isinstance(service_instance, DedicatedCDNServiceInstance)`

broker/tasks/sns.py
```diff
@@ -0,0 +1,71 @@
...
+
+@huey.retriable_task
+def create_notification_topic(operation_id: int, **kwargs):
+    operation = db.session.get(Operation, operation_id)
+    service_instance = operation.service_instance
+
+    operation.step_description = "Creating SNS notification topic"
+    flag_modified(operation, "step_description")
+    db.session.add(operation)
+    db.session.commit()
```
bengerman13, line 24: (not for this PR)
this pattern is repeated in almost every task, should we refactor it out?

bengerman13 (approved):

broker/lib/cdn.py
markdboyd: Absolutely no reason. Is there an inherent benefit to one approach vs the other?

bengerman13: `isinstance` is a builtin, so I think it's a bit more idiomatic, and probably faster (although speed isn't really critical here - nobody would notice if this took seconds to perform, except during testing)

PR #389, opened 2024-09-21
chore: refactor setting step description, etc, to common code

bengerman13:

- centralize where + how we set step description
- fix some alerts from flake8

## Things to check

## Security considerations

None


markdboyd (dismissed):

broker/tasks/cloudwatch.py
```diff
@@ -47,17 +39,80 @@ def create_health_check_alarms(operation_id: int, **kwargs):
     db.session.commit()
 
 
-@huey.retriable_task
-def delete_health_check_alarms(operation_id: int, **kwargs):
-    operation = db.session.get(Operation, operation_id)
+@pipeline_operation("Updating Cloudwatch alarms for Route53 health checks")
+def update_health_check_alarms(operation_id: int, *, operation, db, **kwargs):
```
markdboyd, line 43: This task should be removed

PR #394, opened 2024-09-25
fix: improve cert cleanup for certs that are not in iam

bengerman13:

- check for cert before deleting instead of relying on ignoring a failure we won't get 
-
-

## Things to check

## Security considerations

None


broker/duplicate_certs.py
```diff
@@ -5,6 +5,7 @@
 from broker.aws import alb, iam_govcloud
 from broker.extensions import config, db
 from broker.models import ALBServiceInstance, DedicatedALBServiceInstance, Certificate
+from broker.tasks.iam import _delete_server_certificate
```
bengerman13, line 8: N.B. this import feels lazy, but reorganizing code for one ugly (but functional) import seems like overkill

markdboyd (approved):
Thanks Ben!

PR #407, opened 2025-02-27
allow updating instances that reference cache policy ids

bengerman13:

- allow updating instances that reference cache policy ids

## Things to check

## Security considerations

None - this doesn't change the security posture of this app or the instances it touches.


pburkholder (approved):
You misspelled "difference" in a comment, but LGTM

PR #408, opened 2025-03-14
Refactor dedicated ALB logic to use mapping for ALB listener to org

markdboyd:
Related to https://github.com/cloud-gov/private/issues/1395

- remove DEDICATED_ALB_LISTENER_ARNS config property
- add new DEDICATED_ALB_LISTENER_ARN_MAP config property for map between ALB listener ARNs and dedicated orgs
- update config tests
- Update DedicatedALBListener.load_albs to use new DEDICATED_ALB_LISTENER_ARN_MAP config property and update unit tests
- Remove logic in `get_potential_listeners_for_dedicated_instance` to select an ALB listener with no associated org, since we now assume that all listeners are explicitly mapped to an associated org. Remove tests that verified selection of ALB listeners with no associated org
- Update duplicate cert removal logic to use new DEDICATED_ALB_LISTENER_ARN_MAP property and update tests

## Things to check

## Security considerations

Having an explicit and stable mapping between an org and a dedicated ALB allows for more targeted security measures


bengerman13:
I think this all looks reasonable, but do we need a data migration to make sure we don't have any un-assigned dedicated albs?

PR #416, opened 2025-04-01
Fix CDN update distribution logic

markdboyd:

The `update_distribution` logic was not updating CDN with dedicated WAF instances with the correct tags nor applying the dedicated WAF web ACL to the CDN itself. This PR fixes both of those issues and adds the relevant tests for the updated behaviors.

## Things to check

## Security considerations

The changes in the PR themselves have no security impact as they are just bug fixes and do not expose anything sensitive. For users of CDN with dedicated WAF services though, the security benefit of these changes is actually getting a dedicated WAF for your CDN.

### DIFF
--- a/broker/tasks/cloudfront.py
+++ b/broker/tasks/cloudfront.py
@@ -79,7 +79,6 @@ def is_cdn_with_dedicated_waf_instance(service_instance) -> bool:
 
 
 def update_cdn_with_dedicated_waf_instance_tags(tags: list[Tag]) -> list[Tag]:
-    # tags = service_instance.tags if service_instance.tags else []
     tags = add_tag(tags, "has_dedicated_acl", "true")
     return tags
 
--- a/broker/tasks/cloudfront.py
+++ b/broker/tasks/cloudfront.py
@@ -3,7 +3,7 @@ import time
 
 from broker.aws import cloudfront
 from broker.extensions import config
-from broker.lib.tags import add_tag
+from broker.lib.tags import add_tag, Tag
 from broker.models import CDNServiceInstance, CDNDedicatedWAFServiceInstance
 from broker.tasks.huey import pipeline_operation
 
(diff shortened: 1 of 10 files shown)


broker/tasks/cloudfront.py
```diff
@@ -71,6 +71,18 @@ def get_custom_error_responses(service_instance):
...
 
 
+def is_cdn_with_dedicated_waf_instance(service_instance) -> bool:
+    return (
+        isinstance(service_instance, CDNDedicatedWAFServiceInstance)
+        and service_instance.dedicated_waf_web_acl_arn
+    )
+
+
+def update_cdn_with_dedicated_waf_instance_tags(tags: list[Tag]) -> list[Tag]:
```
bengerman13, line 81: (opinion) this feels a little _too_ DRY, at least as a bare function. Should it be an instance method instead?

broker/tasks/cloudfront.py
```diff
@@ -332,12 +341,28 @@ def update_distribution(operation_id: str, *, operation, db, **kwargs):
...
     config["CustomErrorResponses"] = get_custom_error_responses(service_instance)
 
+    tags = service_instance.tags if service_instance.tags else []
+
+    if (
+        isinstance(service_instance, CDNDedicatedWAFServiceInstance)
+        and service_instance.dedicated_waf_web_acl_arn
+    ):
+        config["WebACLId"] = service_instance.dedicated_waf_web_acl_arn
+        tags = update_cdn_with_dedicated_waf_instance_tags(tags)
```
bengerman13, line 351: won't this blow up if the service instance already has the `has_dedicated_acl` tag?

markdboyd: Very good point. I updated the logic to handle that case and I added a test to catch it

bengerman13 (approved):

Issue #412, opened 2025-03-28
Support managed cache and viewer policies for CDNs


markdboyd:
CloudFront offers [several very useful managed cache policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html) and [managed viewer policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-origin-request-policies.html)

These policies offer simpler and more powerful cache/viewer configurations than the existing "legacy" configurations, so we should offer them to our customers

## Security considerations

I do not think any of these managed policies would pose a security risk to our customers

## Implementation sketch

* [x] Add support for specific a custom cache policy for CDN services. Only the AWS managed policies should be supported options
* [x] Add tests for new custom cache policy behavior
* [x] Add support for specific a custom viewer policy for CDN services. Only the AWS managed policies should be supported options
* [x] Add tests for new custom viewer policy behavior
* [x] Add post to website to announce new features. Update external-domain-broker documentation page to document new options

markdboyd:
I manually tested by creating a CDN in production with the `cache_policy` and `origin_request_policy` parameters and confirmed that they were applied successfully

PR #418, opened 2025-04-24
Support managed cache viewer policies

markdboyd:

Part of https://github.com/cloud-gov/external-domain-broker/issues/412

This PR adds support for users specifying [an AWS managed cache policy](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html) when creating or updating an external-domain CDN service.

[According to the AWS documentation, specifying ForwardedValues explicitly is deprecated](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_DefaultCacheBehavior.html#cloudfront-Type-DefaultCacheBehavior-ForwardedValues), so at some point the broker needs to support setting cache policies directly.

- Refactor some of the helper functions in `broker/api.py` into separate files: `broker/lib/alb.py`, `broker/lib/cdn.py`, `broker/lib/client_error.py` for easier code scanning and maintenance
- Add `cache_policy_id` and `origin_request_policy_id` to `CDNServiceinstance` model
- Specify allowed cache policies in `config.py`. The allowed values are a subset of the [full set of allowed values in AWS](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html), as some of the policies provided by AWS don't make sense for our platform.
- Update `create_distribution` and `update_distribution` to set `CachePolicyId` if one is specified. If a cache policy is specified `ForwardedValues` are also removed (and vice versa), because these properties are mutually exclusive: https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_DefaultCacheBehavior.html#cloudfront-Type-DefaultCacheBehavior-ForwardedValues
- Add `parse_cache_policy` for parsing the supplied cache policy name from create/update params, if one was provided. The function raises an error if the policy is not in the set of allowed values
- Add API tests for CDN create/update operations to ensure that cache policies are set appropriately when provided
- Refactor `_distribution_config_with_tags ` in `fake_cloudfront.py` to call `_distribution_config`, which reduces a lot of duplicate code
- Refactor methods for tags on `CDNDedicatedWAFServiceInstance` and `expect_tag_resource` call mocker to clean up test boilerplate

## Things to check

## Security considerations

The code changes themselves are not sensitive. Allowing users to choose a cache policy via the broker should have no negative effect on security, though users making a poor choice of caching policy could make their CDN caching less effective. At the same time, the managed policies offered by AWS may likely provide much more effective caching than the existing configuration with `ForwardedValues`.


broker/api.py
```diff
@@ -24,9 +24,26 @@
...
 
 from broker import validators
+from broker.aws import cloudfront
 from broker.extensions import config, db
-from broker.lib.cdn import is_cdn_instance, is_cdn_dedicated_waf_instance
+from broker.lib.alb import (
+    validate_migration_to_alb_params,
+    update_alb_params_for_migration,
+)
+from broker.lib.cache_policy_manager import CachePolicyManager
```
bengerman13, line 33: pulling `cloudfront` and `CachePolicyManager` in here feels like we're mixing abstraction levels with the rest of the `api`
this is very much opinion, and I'm having a hard time explaining why it feels like the wrong level or how to fix it, so feel free to ignore

broker/lib/cache_policy_manager.py
```diff
@@ -0,0 +1,29 @@
...
+        self.cloudfront = cloudfront
+
+    def get_managed_cache_policies(self) -> dict[str, str]:
+        policy_type = "managed"
+        if policy_type not in self.policies:
+            self._list_cache_policies(policy_type)
+        return self.policies[policy_type]
+
+    def _list_cache_policies(self, policy_type):
+        # TODO: do we need to handle paging?
```
bengerman13, line 16: `boto3` often (but not always) handles paging under the hood

broker/lib/cache_policy_manager.py
```diff
@@ -0,0 +1,29 @@
...
+class CachePolicyManager:
+    def __init__(self, cloudfront):
+        self.policies = {}
+        self.cloudfront = cloudfront
+
+    def get_managed_cache_policies(self) -> dict[str, str]:
+        policy_type = "managed"
+        if policy_type not in self.policies:
+            self._list_cache_policies(policy_type)
+        return self.policies[policy_type]
```
bengerman13, line 13: if I'm reading the intent here correctly, we want `policies` to be a lazy-loaded property, and we're testing that it's loaded by looking for `"managed"` in the dict because we expect it to always be there. If that's the case, we can do this a little more idiomatically with something like:

```python
def CachePolicyManager:
    def __init__(self, cloudfront):
        self._policies = None
        self.cloudfront = cloudfront

    @property
    def policies(self) -> dict[str, str]:
        if self._policies is None:
            self._policies = self._list_cache_policies()
        return self._policies
```

(although now that I've typed that all out, I see that we're only ever getting `"managed"` policies, so I think I'm missing something here)

broker/lib/cdn.py
```diff
@@ -12,3 +27,188 @@ def is_cdn_instance(service_instance) -> bool:
...
+def parse_cache_policy(params, cache_policy_manager: CachePolicyManager) -> str:
+    cache_policy = params.get("cache_policy", None)
+    if not cache_policy:
+        return None
+    if cache_policy not in config.ALLOWED_AWS_MANAGED_CACHE_POLICIES:
+        raise errors.ErrBadRequest(
+            f"'{cache_policy}' is not an allowed value for cache_policy."
+        )
+    managed_cache_policies = cache_policy_manager.get_managed_cache_policies()
+    if cache_policy in managed_cache_policies.keys():
```
bengerman13, line 48: `x in my_dict` is the same thing as `x in my_dict.keys()`
```suggestion
    if cache_policy in managed_cache_policies:
```

broker/lib/cdn.py
```diff
@@ -12,3 +27,188 @@ def is_cdn_instance(service_instance) -> bool:
...
+        return None
+
+    return params.get("alarm_notification_email")
+
+
+def parse_cache_policy(params, cache_policy_manager: CachePolicyManager) -> str:
+    cache_policy = params.get("cache_policy", None)
+    if not cache_policy:
+        return None
+    if cache_policy not in config.ALLOWED_AWS_MANAGED_CACHE_POLICIES:
```
bengerman13, line 43: this is a duplicate of the test we run when loading the cache policies

broker/lib/cdn.py
```diff
@@ -12,3 +27,188 @@ def is_cdn_instance(service_instance) -> bool:
...
+        raise errors.ErrBadRequest(
+            f"'{cache_policy}' is not an allowed value for cache_policy."
+        )
+    managed_cache_policies = cache_policy_manager.get_managed_cache_policies()
+    if cache_policy in managed_cache_policies.keys():
+        return managed_cache_policies[cache_policy]
```
bengerman13, line 49: I think that de-duplicating the logic from CachePolicyManager, this can be simplified to: 
```suggestion
def parse_cache_policy(params, cache_policy_manager: CachePolicyManager) -> str:
    cache_policy = params.get("cache_policy", None)
    return cache_policy_manager.get(cache_policy, None)
```

bengerman13 (approved):

broker/lib/cdn.py
markdboyd: 6960f98d2286aaa9d3c31040a71bdb8775918d48

broker/lib/cdn.py
markdboyd: 6960f98d2286aaa9d3c31040a71bdb8775918d48

broker/lib/cache_policy_manager.py
markdboyd: I implemented the feedback: 6960f98d2286aaa9d3c31040a71bdb8775918d48. Thoughts?

You're right that currently we're only handling `managed` policies, but I'm future-proofing a little bit in case when we want to have this same code handle `custom` policies

broker/lib/cdn.py
markdboyd: Hmm, yeah, it is, but I do like the idea of throwing a meaningful error to users if they specify a non-supported value rather than just silently doing nothing. Thoughts?

broker/api.py
markdboyd: I agree. This case is somewhat special because we need to talk to AWS to get the policy IDs so that we can set the ID for the specified policy on the instance around here:

https://github.com/cloud-gov/external-domain-broker/blob/6960f98d2286aaa9d3c31040a71bdb8775918d48/broker/lib/cdn.py#L94

And that needs to happen in the context of the API request.  Still thinking of how to change it

I could do this in `broker/extensions`

```
cache_policy_manager = CachePolicyManager(cloudfront)
```

That's just moving things around though

bengerman13: I think that move might be worthwhile still, because then we're keeping `cloudfront` itself out of the `api` scope

broker/lib/cache_policy_manager.py
bengerman13: ah, yeah, that looks a lot better :+1:

bengerman13 (approved):

PR #421, opened 2025-05-01
Fix policy handling

markdboyd:

When running the acceptance tests from #420 in staging, we are seeing this error:

```
KeyError: 'CachingDisabled'
```

Further investigation revealed that for [a policy name of `CachingDisabled` as mentioned in the AWS documentation](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html), the actual policy name returned by the CloudFront ListCachePolicies API is `Managed-CachingDisabled`. Since the broker was caching the policies from in AWS in a hash keyed by the policy name in the API response, `CachingDisabled` was not found and hence the `KeyError`. 

The easiest solution is to have the broker expect as input the names of the policies as returned by the CloudFront API. I will make sure to document these values on the external-domain-broker service page on the website.

I tested and discovered that this same problem would have occurred for the origin request policy names, since their names also differ between the AWS documentation page and the API response. So I fixed the problem there again by expecting the input policy names to match the values in the API response.

Most of the changes in this PR are just about updating test fixtures and expectations to match policy names as they come back in the AWS CloudFront API responses.

- Update config for allowed cache policies and origin request policies to use policy names as returned by CloudFront API
- Update various unit tests to expect policy names as returned by the CloudFront API
- Add `is_cache_policy_allowed()` function for testing if a policy name is allowed by the broker
- Update `parse_cache_policy()` to use `is_cache_policy_allowed()` for testing if the given policy name is allowed by the broker
- Add `is_origin_request_policy_allowed()` function for testing if a policy name is allowed by the broker
- Update `parse_origin_request_policy()` to use `is_origin_request_policy_allowed()` for testing if the given policy name is allowed by the broker
- Update `test_cloudfront_update_distribution_sets_cache_policy` test to ensure that existing value for `DefaultCacheBehavior['Compress']` does not get inadvertently deleted on update

## Things to check

## Security considerations

None. These code changes are not sensitive.

### DIFF
--- a/broker/config.py
+++ b/broker/config.py
@@ -62,10 +62,10 @@ class Config:
         self.REQUEST_TIMEOUT = 30
 
         # see https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html
-        # NOTE: The values in the list are not the same as the policy names on the AWS documentation. These values
-        # are the actual policy names as returned by a CloudFront ListCachePolicies request. Since our code will
-        # check for allowed policies against the list of names returned by the API, we use the names from the API
-        # response instead of the documentation.
+        # NOTE: The values in the list are not the same as the policy names on the AWS documentation page. These
+        # values are the actual policy names as returned by a CloudFront ListCachePolicies request. Since our code
+        # will check for allowed policies against the list of names returned by the API, we use the names from the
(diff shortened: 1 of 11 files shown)


JasonTheMain (approved):
Looks good to me

PR #423, opened 2025-05-02
Fix apply cache policy logic

markdboyd:

The acceptance tests to create a CDN with dedicated WAF service instance using a managed cache policy and managed origin request policy are failing with

```
The parameter MinTTL cannot be used when a cache policy is associated to the cache behavior
```

Upon reviewing the [relevant AWS documentation](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_DefaultCacheBehavior.html), it seems like the `MinTTL`, `MaxTTL`, and `DefaultTTL` properties are deprecated when you are supply a cache policy. 

So this PR updates the CDN provisioning logic to unset those properties when a `CachePolicyId` is supplied.

The PR also updates the acceptance test to use the `Managed-CachingOptimized` cache policy.

## Things to check

## Security considerations

None. The code changes are not sensitive

### DIFF
--- a/tests/lib/fake_cloudfront.py
+++ b/tests/lib/fake_cloudfront.py
@@ -507,6 +507,9 @@ class FakeCloudFront(FakeAWS):
             )
         else:
             default_cache_behavior.update({"CachePolicyId": cache_policy_id})
+            default_cache_behavior.pop("DefaultTTL", None)
+            default_cache_behavior.pop("MinTTL", None)
+            default_cache_behavior.pop("MaxTTL", None)
 
         if origin_request_policy_id:
             default_cache_behavior.update(
--- a/acceptance/run.sh
+++ b/acceptance/run.sh
@@ -160,7 +160,7 @@ tests() {
       "$SERVICE_NAME" \
       "$PLAN_NAME" \
       "$INSTANCE" \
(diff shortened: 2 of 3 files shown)


tests/lib/fake_cloudfront.py
```diff
@@ -507,6 +507,9 @@ def _distribution_config(
             )
         else:
             default_cache_behavior.update({"CachePolicyId": cache_policy_id})
+            default_cache_behavior.pop("DefaultTTL", None)
```
pburkholder, line 510: I'm puzzled why you're popping out this values here when in https://github.com/cloud-gov/external-domain-broker/pull/423/files#diff-1eee4c4df9a8538fae63c404bedd2d08d98360b41a6c6be6e29ac1c1a5b4c0efR102 you say these are set by the CachePolicy.

markdboyd: It is a bit confusing. Basically, I think when you supply a `CachePolicyId`, the managed cache policy sets the values of `MinTTL`, `DefaultTTL`, and `MaxTTL` automatically, so there's a conflict if you're setting those values yourself.

That's why [this documentation](https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_DefaultCacheBehavior.html)says those properties are deprecated when you use cache policies.

And [the documentation on managed cache policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html) outlines the default values that cache policies will set for these properties

PR #424, opened 2025-05-07
Fix update CDN policy logic to preserve existing values

markdboyd:

- [fix update_default_cache_behavior() to preserve managed cache policy and managed origin request policyfrom existing distribution config, even if no updates are specified to these values](https://github.com/cloud-gov/external-domain-broker/commit/bec2a8220d969f64ddf375f9e0a8ab3001d3feb7) 
- [add unit test to confirm that CDN updates preserve existing CDN configuration](https://github.com/cloud-gov/external-domain-broker/commit/90571fdfb4850991e47146f8f89bd38b77bd2825) 

## Things to check

## Security considerations

None. These code changes are not sensitive


JasonTheMain (approved):
Looks good to me

PR #430, opened 2025-05-27
Support update from dedicated ALB to dedicated CDN plan

markdboyd:

This PR adds the code necessary to support a direct upgrade from `domain-with-org-lb` service instances to `domain-with-cdn-dedicated-waf` service instances.

- Add pipeline for migrating from `domain-with-org-lb` service instances to `domain-with-cdn-dedicated-waf` service instances with no downtime.
- Add `MigrateDedicatedALBToCDNDedicatedWafServiceInstance` model for handling transition from `domain-with-org-lb` services to `domain-with-cdn-dedicated-waf` to services
- Added tasks:
  -  `remove_alb_certificate_during_update_to_cdn_dedicated_waf`
  - `store_alb_certificate`
- Various refactors to support testing

## Things to check

## Security considerations

None. The changes are not sensitive


tests/lib/alb/provision.py
```diff
@@ -100,9 +106,11 @@ def subtest_provision_creates_provision_operation(
     return operation_id
 
 
-def subtest_provision_provisions_ALIAS_records(tasks, route53, instance_model):
+def subtest_provision_provisions_ALIAS_records(
+    tasks, route53, instance_model, service_instance_id="4321"
```
JasonTheMain, line 110: i've seen you pass this a few times. is there a reason you do it this way instead of using a global or etc? 
service_instance_id="4321"

JasonTheMain (approved):
Looks good to me

PR #435, opened 2025-07-09
Provision dedicated WAF for dedicated ALBs

markdboyd:
## Background

This broker includes a `domain-with-org-lb` plan. Instances using this plan will have certificates for their domain(s) attached to a load balancer specified for their organization. There may be multiple load balancers specified for the current organization, so the broker chooses the one which has the least certificates provisioned on it. [There is a limit of 25 certificates for a load balancer listener.](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html)

The load balancers themselves are provisioned elsewhere. This broker is aware of them through the `DEDICATED_ALB_LISTENER_ARN_MAP` environment variable that is specified in the pipeline when the app is deployed.

This PR updates the broker to provision a dedicated WAF web ACL for a load balancer used by `domain-with-org-lb` instances.

The PR provides two ways to provision these dedicated WAF web ACLs:

- Adds a `create_cdn_web_acl` task. For a given `domain-with-org-lb` service, the task will look up the associated LB, determine if there is already a web ACL associated, and provision one if not.
- Adds a  `add_dedicated_alb_waf_web_acls` command that can be invoked directly on the application. This command is necessary so that we can backfill existing load balancers for `doman-with-org-lb` with WAF web ACLs without requiring customers to update their services.

There are a number of code changes made in support of the broader feature change:

- Added a `DedicatedALB` model to track the current load balancers used by `domain-with-org-lb` instances
  - Each `alb_arn` value in this table must be unique
  - This model will also track whether a WAF web ACL is associated with the load balancer
  - Made the `alb_arn` column of the `DedicatedALBListener` model a foreign key reference to this table 
- Added a `create_cdn_web_acl` task to provision a WAF web ACL on load balancer for `domain-with-org-lb` instances
- Added a `put_alb_waf_logging_configuration` task to put logging configuration a WAF web ACL on load balancer for `domain-with-org-lb` instances
- Updated these operations to create WAF web ACLs on load balancers for `domain-with-org-lb` instances:
  - `domain-with-org-lb` provision
  - `domain-with-org-lb` update
  - `domain` -> `domain-with-org-lb` update
- Tests (unit and integration) were added for all of the new or changed behaviors
- Updated package versions. This did not turn out to be strictly necessary, but since it is good code hygiene, I left it in
  
There is various code refactoring and reorganization as part of these changes.

This PR **does not**:

- Add provisioning/deprovisioning of load balancers for `domain-with-org-lb` instances
  - This would have increased the scope of the work and code changes significantly
- Delete WAF web ACLs when deprovisioning `domain-with-org-lb` instances
  - The only appropriate time to remove the WAF web ACL for a load balancer would be just before deprovisioning it. Since the load balancers are not provisioned/deprovisioned by this broker, it did not make sense to include this functionality here

## Things to check

## Security considerations

Giving customers using `domain-with-org-lb` instances a dedicated WAF for their org is beneficial for security. Also the package updates for the broker dependencies are good for security and may address vulnerabilities.


pyproject.toml
```diff
@@ -31,6 +31,7 @@ ignore_patterns = ["venv/*", ".venv/*"]
 addopts = ["-vv"]
 filterwarnings = [
   'error',
-  'ignore:X509Extension support in pyOpenSSL is deprecated. You should use the APIs in cryptography.:DeprecationWarning',
-  'ignore:CSR support in pyOpenSSL is deprecated:DeprecationWarning'
+  'ignore:TLSALPN01Server is deprecated:DeprecationWarning',
+  'ignore:TLSALPN01Response is deprecated:DeprecationWarning',
+  'ignore:TLSALPN01 is deprecated:DeprecationWarning'
```
markdboyd, line 36: These deprecation warnings are a byproduct of upgrading `acme` to `> 4.1`. As far as I can tell, we are not actually using TLSALPN01 challenges, but the code is written to emit these deprecation warnings regardless. 

All of the unit/integration tests still pass. And I'll verify that the acceptance tests pass in CI

PR #443, opened 2025-09-05
Update associate WAF web ACL logic

markdboyd:

Follow-up to #442. When updating a dedicated org load balancer to **one WAF web ACL to be shared by all load balancers for an organization, rather than one WAF web ACL for each of the load balancers**, we need to clean up the old WAF web ACL that was shared by all the load balancers.

This PR implements the logic to delete the obsolete WAF web ACL once the new one is associated.

- Rename `associate-dedicated-alb-waf-web-acls` to `update-dedicated-alb-waf-web-acls`
- Update logic for `update-dedicated-alb-waf-web-acls` to:
  - Associate new web ACL tracked by database only if it is not already associated with the ALB
  - Wait to confirm that new web ACL has been associated with ALB
  - Delete the obsolete web ACL no longer associated with the ALB
- Update tests for new behavior

## Things to check

## Security considerations

None. There is no change in the behavior of the web ACLs, this is just utility code for handling the creation and association of new web ACLs with dedicated ALBs.

### DIFF
--- a/broker/app.py
+++ b/broker/app.py
@@ -22,7 +22,7 @@ from broker.commands.duplicate_certs import (
 from broker.commands.tags import add_dedicated_alb_tags
 from broker.commands.waf import (
     create_dedicated_alb_waf_web_acls,
-    associate_dedicated_alb_waf_web_acls,
+    update_dedicated_alb_waf_web_acls,
 )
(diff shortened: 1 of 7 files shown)


JasonTheMain (approved):
Looks good to me

PR #445, opened 2025-09-10
Fix create dedicated ALB WAF command

markdboyd:

- Update `waf.create_web_acl` to return a boolean indicating whether the requested web ACL already exists (True) or not (False)
- Update `create_dedicated_alb_waf_web_acls` command to get and set the web ACL info on the dedicated ALB if the web ACL already exists
- Add test of creating web ACL for multiple dedicated ALBs with the same dedicated org
- Refactor/update tests to match new behavior

## Things to check

## Security considerations

None, just fixing bug when attempting to create web ACL for multiple dedicated ALBs with the same dedicated org

### DIFF
--- a/tests/integration/commands/test_waf.py
+++ b/tests/integration/commands/test_waf.py
@@ -1,5 +1,4 @@
 import pytest
-import random
 
 from broker.commands.waf import (
     create_dedicated_alb_waf_web_acls,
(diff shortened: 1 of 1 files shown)


JasonTheMain (approved):
Looks good to me

PR #480, opened 2026-06-22
replace deprecated code for generating a private key

markdboyd:

- replace deprecated code for generating a private key

## Things to check

## Security considerations

Replacing deprecated code with more well maintained code is good for security


JasonTheMain (approved):
LGTM

